PR #132, opened 2026-04-08
feat(metal): set cull mode and front face winding

jdbann:
For Metal, cull mode and front face winding are set on the render pass rather than the render pipeline. Store the raw values when creating RenderPipeline and set the state as part of SetPipeline.

### DIFF
--- a/hal/metal/device.go
+++ b/hal/metal/device.go
@@ -658,7 +658,13 @@ func (d *Device) CreateRenderPipeline(desc *hal.RenderPipelineDescriptor) (hal.R
 	if pl, ok := desc.Layout.(*PipelineLayout); ok {
 		pipeLayout = pl
 	}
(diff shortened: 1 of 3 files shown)


kolkov (approved):
Clean, minimal, correct. Metal API requires cull mode and front face winding to be set on the render encoder (not pipeline state), and this PR follows that architecture properly. Conversion functions already existed in conv.go — this just wires them in.

All other backends (Vulkan, DX12, GLES, Software) already handle cull mode correctly. This was the only gap.

Verified: no matching issue on other backends.

jdbann:
Thank you! Let me know if there are any other open Metal issues that need attention 👍

kolkov:
https://github.com/gogpu/wgpu/issues/133 - This isn't Metal, but it's a software backend. Try it on macOS. See if it works, and if not, see if you can fix it.
Thanks!

PR #136, opened 2026-04-08
feat(metal): support depth textures in render pass

jdbann:
In `hal/metal/encoder.go`, `BeginRenderPass` correctly handles depth attachments. `MTLRenderCommandEncoder` does not get depth stencil state from the render pipeline - instead it is set on the encoder directly via `setDepthStencilState(_:)` and `setDepthBias(_:slopeScale:clamp:)`.

To enable this, `MTLDepthStencilState` and bias values are stored on `RenderPipeline` when created and set as part of `SetPipeline`.

I have verified basic depth texture use locally but I'm not familiar enough with the depth bias values to know how to verify their behaviour. Stencil support is not included here for the same reason.

### DIFF
--- a/hal/metal/device.go
+++ b/hal/metal/device.go
@@ -624,6 +624,24 @@ func (d *Device) CreateRenderPipeline(desc *hal.RenderPipelineDescriptor) (hal.R
 		}
 	}
 
+	var depthStencilState ID
+	var depthBias, depthSlopeScale, depthClamp float32
+	if desc.DepthStencil != nil {
+		// Set depth attachment format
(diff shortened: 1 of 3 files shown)


kolkov (approved):
Clean depth/stencil implementation following the same pattern as #132. Matches Vulkan pipeline depth state. CI green on all platforms.

kolkov:
@jdbann Thanks! Depth textures merged and shipped in wgpu v0.24.6 together with your cull mode fix from #132. Two Metal improvements in one release — great work!

PR #149, opened 2026-04-24
Particle simulator example

snakeru:
Hi there!

I am finally done with my simulation example. It currently does almost exactly what was originally planned. There is only one feature missing: when there are way too many particles - some of them might become frozen. In my Go code I had a workaround for that, but writing shaders requires a different mindset so that one was dropped. But I believe that it doesn't really hurt the example too much - its goal is to look interesting, not address the actual physics use-cases.

Effectively there are only 3 important files (all 3 wgsl), everything else is just a harness. Please excuse me for the less-than-beautiful go code - I was lazy enough to just keep it barely in shape for the rest to work. For example there are 4 command encoders currently. I have quickly tried to use a single one but my knowledge of WGPU is lacking.

Thank you for your beautiful project! Let me know if you want me to make any corrections to the code.


kolkov (commented):
Thanks @snakeru! Great to see the particle simulator from wgpu#120 and gg#228 land as a proper example. We tested it locally -- 60 FPS stable on Intel Iris Xe Vulkan, looks beautiful.

A few things to address before merging:

### Required

**1. License: AGPL -> MIT**

The WGSL shaders have `SPDX-License-Identifier: AGPL-3.0-only`. The wgpu repository is MIT-licensed. All code contributed must be MIT. Please change to:

```wgsl
// Copyright 2026 The GoGPU Authors
// SPDX-License-Identifier: MIT
```

This matches our existing examples (see `examples/compute-particles/main.go`).

**2. go.mod module path**

`module main` should follow Go conventions for examples inside a larger module. Our other examples don't use separate go.mod files -- they compile as part of the main module. Please remove the separate `go.mod`/`go.sum` and use the wgpu module's root go.mod instead:

```go
// main.go
package main

import (
    "github.com/gogpu/gogpu"
    "github.com/gogpu/wgpu"
    // ...
)
```

If a separate go.mod is needed (e.g., for the gogpu dependency), the module path should be descriptive: `module github.com/gogpu/wgpu/examples/compute-and-render-particles`.

**3. Remove hardcoded exit**

```go
if frameCounter > 163900 {
    log.Fatal("done")
}
```

This looks like a leftover from debugging. Examples should run until the user closes the window.

**4. Uniform struct mismatch across shaders**

Two different `UniformData` layouts share one buffer:

- `bucketer.wgsl` / `interactor.wgsl`: `{ particleCount: u32, gridWidth: u32, gridHeight: u32 }` (12 bytes)
- `bouncer.wgsl` / `render.wgsl`: `{ particleCount: u32, pad1: u32, pad2: u32, gridWidth: f32, gridHeight: f32 }` (20 bytes)

The Go code writes 20 bytes. The bucketer/interactor read `gridWidth` at offset 4 as `u32`, but the Go code writes the `u32` grid dimensions there, so it works by coincidence (integer values 16 and 9). However, this is fragile -- a single unified struct across all shaders would be cleaner and safer.

### Suggested improvements (non-blocking)

**5. Single command encoder**

You mentioned trying a single encoder. Here is the pattern:

```go
enc, _ := dev.CreateCommandEncoder(nil)

// Compute pass 1
cp1, _ := enc.BeginComputePass(nil)
cp1.SetPipeline(bucketerPipe)
cp1.SetBindGroup(0, bucketerBG, nil)
cp1.Dispatch(...)
cp1.End()

// Compute pass 2
cp2, _ := enc.BeginComputePass(nil)
cp2.SetPipeline(interPipe)
cp2.SetBindGroup(0, interBG, nil)
cp2.Dispatch(...)
cp2.End()

// Compute pass 3 + Render pass -- all in one encoder
// ...

cmds, _ := enc.Finish()
dev.Queue().Submit(cmds)  // single submit
```

One encoder, one submit -- reduces driver overhead and lets the GPU pipeline the passes.

**6. Per-frame allocation**

```go
cellData := make([]byte, cellBufSize) // allocates every frame
```

Move `cellData` to a field on `sim` and reuse it with `clear(cellData)` (Go 1.21+).

**7. Script file**

The `run` file (no extension) should be `run.sh` with a proper shebang (`#!/bin/bash`).

**8. Copyright header on main.go**

Add the standard header to match other examples:
```go
// Copyright 2026 The GoGPU Authors
// SPDX-License-Identifier: MIT
```

---

Overall this is a solid compute + render example. The 3-stage pipeline (bucketer -> interactor -> bouncer) with spatial hashing is a great showcase of what wgpu compute shaders can do. Looking forward to the updated version!

snakeru:
Hi @kolkov,

Thank you for the detailed feedback. Let's discuss the main issue first, everything else is easy to fix.

**License**: unfortunately, I prefer all my public work to be strong copyleft. That's why I usually use something like GPL3 or AGPL - not even LGPL. To my knowledge MIT allows the code to be made proprietary - this is just a plain "no" from me.

That said, my understanding of the purpose of this software was to demonstrate the Go code and abilities of your library. The actual shader code is irrelevant for that purpose - it provides a flashy bait for the user, but the actual meat here is the `.go` file(s). So I insist on using AGPL **only** for the three `.wgsl` files I developed. Everything else could use whatever license - there I can't care less.

Please let me know if you see any way out of this situation. If this is makes this PR unacceptable then I can just publish this project separately in my github account and you can just link to it, this should be next best option.

snakeru:
Actually, I might have found a solution myself that should be good enough. Please stand by.

snakeru:
HI @kolkov 

I think, all points are addressed now, including non-blockers:

## Blockers
1. **LICENSE**: I solved it by publishing this code at github.com/snakeru/wgpu-particles and adding a README.md.This version will become frozen anyway so I can keep working on my AGPL instance while you'll have your MIT-compliant copy. I hope that's fine with you.
2. `go.mod` and `go.sum` are removed. **Please check**, that this code runs for you. For me it doesn't anymore without these.
3. Hard coded exit is removed.
4. Uniform is shortened to have only `u32`. I convert these to `f32` directly in wgsl. It was a premature optimization.

## Non-blockers
5. Single command encoder is added and seems to work. Thank you a lot for the exact recipe!
6. per-frame allocation is removed because the corresponding feature was removed. It was half-baked and need some serious rehaul anyways. While the underlying issue was not very visible in the exact code provided, but as soon as someone cranks up the number of cells (and particles) and/or increases gravitation it starts to show up. The current version does not support as many particles, but does not have this problem either, which I see as a win.
7. `run.sh` now has a proper header
8. copyright header added to he `main.go` as well.

## Further improvements
9. Particles now have different masses/sizes and gravitate towards a central "planet". I think it would be nice to put the WGPU logo in there, but I didn't look into how to do that. I also included a screenshot after a several hours of simulation at 144 FPS.

Please let me know if there are any other changes needed.

snakeru:
Ah, one more remark: at the start of the simulation there are some "density rays" observable for a few seconds. This is caused by insufficient f32 precision. There are several ways to fix this, but I don't currently see it as a big issue.

kolkov:
Hi @snakeru, thanks for addressing all the feedback! The code looks great.

We owe you an apology on the `go.mod` point. In our review we asked you to remove it, and you rightfully warned us: *"Please check, that this code runs for you. For me it doesn't anymore without these."* We should have listened — you were right. The example imports `gogpu`, which is not a dependency of the `wgpu` module, so without a separate `go.mod` it won't compile with `go build ./...` from the wgpu root. Sorry for the back and forth.

After thinking about it more carefully, here are the options — your choice:

**Option A: Restore `go.mod` in the example directory**

Add back `go.mod` with a proper module path:
```
module github.com/gogpu/wgpu/examples/compute-and-render-particles
```
This is the simplest fix. The example will compile standalone and via our `go.work` for local development. The trade-off is that all other wgpu examples are headless (no window), so this would be the only windowed example in the wgpu repo.

**Option B: Move to `gogpu/gogpu` repository**

All our windowed examples (triangle, particles) live in `gogpu/examples/` because they use the gogpu app framework for windowing. Your particle simulator uses `gogpu.NewApp()` and `gogpu.Context`, so architecturally it fits there. You'd open a new PR on `gogpu/gogpu` and we can close this one.

Both options are fine with us. Let us know what you prefer!

snakeru:
I do not have any preference here as this is your project. But from a logical standpoint it probably really belongs to the `gogpu/gogpu`. I can do that tomorrow.

snakeru:
Opened https://github.com/gogpu/gogpu/pull/198

PR #210, opened 2026-06-07
fix/292-gles : Fix gless panic to wayland

lkmavi:


kolkov (approved):
## Review: PR #210 — Fix GLES panic on Wayland (@lkmavi)

Excellent contribution. Three independent fixes, all validated:

### 1. FFI pointer convention fix in `context_linux.go` — CRITICAL

30+ GL calls passed `unsafe.Pointer(&value)` where goffi expected `unsafe.Pointer(&pointer_to_value)`. goffi's `call_unix.go:100` does `*(*uintptr)(avalue[idx])` for PointerType args — it reads 8 bytes to get the pointer value to pass. When we passed `unsafe.Pointer(&uint32_var)`, goffi read 8 bytes from a 4-byte variable → garbage pointer → OpenGL received wrong addresses.

This explains why GLES rendering never worked on Linux — we only tested GLES visually on Windows (WGL + `syscall.SyscallN`, which has different calling semantics). CI integration-gles tests EGL init only, not GL rendering.

**Verified against goffi source:** `internal/arch/amd64/call_unix.go:100`:
```go
case types.PointerType:
    addInt(*(*uintptr)(avalue[idx]))
```

### 2. `RequestAdapterWithSurface` surface hint plumbing — correct

`enumerateDeferredGLES(surfaceHint)` called before `RequestAdapter` ensures GLES gets the real surface. Idempotency via `glesEnumerated` flag prevents double enumeration.

### 3. Nil guard in `Adapter.Open()` — correct defensive fix

Returns descriptive error instead of nil pointer crash.

### 4. EGL `nativeDisplay` for Wayland — correct

Passes `wl_display*` to `GetEGLDisplay` instead of 0 — fixes "second wl_display connection" issue.

### Minor note

`go.mod` moves `goffi` and `x/sys` from indirect to direct — cosmetic, acceptable.

### Tests

202 lines of stub-based tests covering surface hint propagation, idempotency, nil guard. Well structured.

**LGTM. Critical fix for Linux GLES.**

PR #225, opened 2026-06-24
fix(hal): add tanh + integer clamp to SPIR-V interpreter

amery:
## Summary

The software SPIR-V interpreter's GLSL.std.450 dispatch was missing two
common math builtins. Unknown ext-inst opcodes fall through to
`return ValUint(0)`, so shaders calling these got **silently wrong
results (zeros) rather than an error**:

- `tanh()` — `Tanh` (21) absent (all hyperbolics, 19–24, were missing)
- integer `clamp()` — `SClamp` (45) and `UClamp` (44) absent; only the
  float `FClamp` (43) was handled

## Changes

- Implement `Sinh`/`Cosh`/`Tanh`/`Asinh`/`Acosh`/`Atanh` (19–24) and
  `UClamp`/`SClamp` (44–45) in `hal/software/shader/glsl_ext.go`.
- Add `glslTernaryUint` / `glslTernaryInt` helpers, mirroring the
  existing `glslBinary*` ones.
- Interpreter tests for each: hyperbolic rows in `TestGLSLUnaryIntrinsics`,
  plus `TestGLSLSClamp` / `TestGLSLUClamp`.

## Testing

`go test ./hal/software/shader/` and `go vet ./hal/software/shader/` pass.

## Discovery

Surfaced by the born ML framework's webgpu backend on the software path,
where its `Tanh` and `Int32 Clamp` ops returned all-zeros. With these
opcodes implemented, both produce correct results.

No existing issue tracked this; still reproducible on `main`.

### DIFF
--- a/hal/software/shader/glsl_ext.go
+++ b/hal/software/shader/glsl_ext.go
@@ -31,6 +31,13 @@ const (
 	GLSLAcos = 17
 	GLSLAtan = 18
 
+	GLSLSinh  = 19
+	GLSLCosh  = 20
+	GLSLTanh  = 21
+	GLSLAsinh = 22
+	GLSLAcosh = 23
+	GLSLAtanh = 24
+
 	GLSLPow         = 26
 	GLSLExp         = 27
 	GLSLLog         = 28
@@ -46,6 +53,8 @@ const (
 	GLSLUMax   = 41
 	GLSLSMax   = 42
 	GLSLFClamp = 43
+	GLSLUClamp = 44
+	GLSLSClamp = 45
 
 	GLSLFMix       = 46
 	GLSLStep       = 48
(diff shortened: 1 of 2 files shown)


kolkov (approved):
LGTM! Clean implementation, follows all existing patterns.

**Verified:**
- All 8 opcode numbers correct per GLSL.std.450 spec (Sinh=19–Atanh=24, UClamp=44, SClamp=45)
- Hyperbolic ops use `glslUnaryFloat` — consistent with Sin/Cos/Tan pattern
- Integer clamp ops use new `glslTernaryUint`/`glslTernaryInt` helpers — mirrors existing `glslBinary*` style
- Tests cover all branches: saturation boundaries (tanh ±10), domain boundaries (acosh(1)), negative range (SClamp)
- float32↔float64 conversion chain matches every other float op in the file

**One systemic note** (not blocking this PR): unknown GLSL.std.450 opcodes silently return `ValUint(0)` — this is how Born ML got zeros in the first place. Filed internally as BUG-SW-009 to add a warning log for unimplemented opcodes.

Thanks @amery — shipping in v0.30.3.

PR #227, opened 2026-06-25
fix(hal/metal): translate stencil state into MTLDepthStencilState

samyfodil:
### Summary
The Metal HAL never translated a pipeline's stencil state into its
`MTLDepthStencilState`. As a result the stencil test was silently inert on Metal:
any pipeline that relied on it behaved as if `stencilCompareFunction = Always`.

The most visible casualty was the **stencil-then-cover** fill path. The cover pass
draws the path's bounding quad with a `NotEqual`/ref-0 stencil test so only pixels
inside the tessellated path are painted. With the test defaulting to `Always`, the
cover pass painted the **entire bounding quad** instead. On Apple Silicon this meant
rounded UI panels rendered as **squares** and clipped / stencil-masked content
(swatches, toggle pills, slider tracks) **vanished**. Vulkan, D3D12 and GLES already
translate the full stencil state — Metal was the only backend missing it.

### Root cause
`device.go`, `CreateRenderPipeline`: the `MTLDepthStencilDescriptor` was configured
with only `setDepthCompareFunction:` and `setDepthWriteEnabled:`. The
`StencilFront` / `StencilBack` faces, the compare functions, the fail / depth-fail /
pass operations, and the read / write masks from the WebGPU `DepthStencilState` were
never applied, so the descriptor kept Metal's defaults (`Always` / `Keep`).

### Changes
- **`hal/metal/device.go`** — build front and back `MTLStencilDescriptor`s from
  `StencilFront` / `StencilBack` (compare function + fail / depth-fail / pass ops) and
  apply the global read / write masks, then set them on the depth-stencil descriptor.
  WebGPU read/write masks are global (not per-face), so both faces share them.
- **`hal/metal/conv.go`** — add `stencilOperationToMTL`. The WebGPU and Metal
  stencil-op enums use **different numeric values** (e.g. WebGPU `Keep = 1`,
  Metal `Keep = 0`; WebGPU `IncrementWrap = 7`, Metal `IncrementWrap = 6`), so an
  explicit mapping is required — a bit-for-bit reinterpret would be wrong.
- **`hal/metal/conv_test.go`** — table-driven unit test for the new mapping.
- Also `Release` the stencil descriptors (and the depth-stencil descriptor) after
  assignment — the descriptor properties copy, so this fixes a small pre-existing leak.

### Testing
Verified on Apple Silicon (M4) against a real Metal device by driving gg's
stencil-then-cover renderer with GPU→CPU pixel readback (companion regression tests
on the gg side):

- **Circle / rounded rect / even-odd ring** — each shape must mask to its path, not
  flood the cover quad's bounding box.
  - **Before:** bounding-box corners painted opaque (circle rendered as a square,
    even-odd hole filled).
  - **After:** corners transparent, rounded corners clipped, even-odd hole empty.
- `go test ./hal/metal/` passes (excluding a pre-existing, unrelated surface-test
  crash that reproduces on `main`).
- Full gg test tree passes with this change in place.

### Scope / risk
- Additive only; no change to depth-only pipelines (a default/zero stencil face maps
  to `Always` + `Keep`, i.e. pass-through, matching prior behavior).
- Brings Metal in line with the Vulkan / D3D12 / GLES HALs, which already translate
  the full stencil state.

### DIFF
--- a/hal/metal/conv.go
+++ b/hal/metal/conv.go
@@ -225,6 +225,31 @@ func compareFunctionToMTL(fn gputypes.CompareFunction) MTLCompareFunction {
 	}
 }
 
+// stencilOperationToMTL maps a WebGPU stencil operation to Metal. The two enums
+// use different numeric values, so an explicit switch is required.
+func stencilOperationToMTL(op gputypes.StencilOperation) MTLStencilOperation {
+	switch op {
+	case gputypes.StencilOperationKeep:
+		return MTLStencilOperationKeep
+	case gputypes.StencilOperationZero:
+		return MTLStencilOperationZero
+	case gputypes.StencilOperationReplace:
+		return MTLStencilOperationReplace
+	case gputypes.StencilOperationInvert:
+		return MTLStencilOperationInvert
+	case gputypes.StencilOperationIncrementClamp:
+		return MTLStencilOperationIncrementClamp
+	case gputypes.StencilOperationDecrementClamp:
+		return MTLStencilOperationDecrementClamp
+	case gputypes.StencilOperationIncrementWrap:
+		return MTLStencilOperationIncrementWrap
+	case gputypes.StencilOperationDecrementWrap:
+		return MTLStencilOperationDecrementWrap
+	default:
+		return MTLStencilOperationKeep
+	}
+}
+
 // primitiveTopologyToMTL converts WebGPU primitive topology to Metal primitive type.
 func primitiveTopologyToMTL(topology gputypes.PrimitiveTopology) MTLPrimitiveType {
 	switch topology {
--- a/hal/metal/conv_test.go
+++ b/hal/metal/conv_test.go
@@ -273,6 +273,37 @@ func TestCompareFunctionToMTL(t *testing.T) {
 	}
 }
 
+// TestStencilOperationToMTL guards the WebGPU→Metal stencil-op mapping. The two
+// enums use different numeric values (WebGPU Keep=1, Metal Keep=0), so this must
+// be an explicit table — a missing/incorrect entry silently disables stencil ops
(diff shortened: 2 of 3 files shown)


kolkov (approved):
LGTM! Solid fix — brings Metal HAL into parity with Vulkan/DX12/GLES for stencil state.

**Verified against Rust wgpu-hal Metal** (`wgpu-hal/src/metal/device.rs:40-67, conv.rs:244-258`):
- `newStencilFaceDescriptor()` matches Rust's `create_stencil_desc()` — same 6 properties
- `stencilOperationToMTL` switch matches Rust's explicit match — required because WebGPU `Invert=4` ≠ Metal `Invert=5`
- `Release(front)`, `Release(back)`, `Release(depthStencilDesc)` — correct, Metal properties copy. The `depthStencilDesc` release also fixes a pre-existing leak
- Read/write masks global (both faces share) — per WebGPU spec

**One pre-existing gap** (not blocking this PR, existed before your change):

Rust wgpu sets `stencilAttachmentPixelFormat` separately from `depthAttachmentPixelFormat` when the format has stencil aspects (device.rs:1453-1458):
```rust
if aspects.contains(FormatAspects::STENCIL) {
    descriptor.set_stencil_attachment_pixel_format(raw_format);
}
```

Our `device.go:682` only sets `setDepthAttachmentPixelFormat:` — missing `setStencilAttachmentPixelFormat:` for formats like `Depth24PlusStencil8`. Without it, Metal may not configure stencil attachment on the render pipeline. This would be a natural follow-up if you want to tackle it — otherwise we'll track it separately.

Thanks @samyfodil — great work, as always.

lkmavi (approved):
Great job! The code looks very good.

PR #239, opened 2026-07-07
fix(metal): live-resize IOSurface pool churn, retain leak, and test crash

lkmavi:
**Summary:**

Addresses a 4+ GB memory accumulation during macOS live window resize, a
per-frame IOSurface pin visible under any resize, and a pre-existing test crash
in `hal/metal`.

**Root causes fixed:**

1. **Retain leak** — `AcquireTexture` called `Retain` on both the drawable and
   the MTLTexture. `DestroyTexture` skipped `isExternal` textures; nothing else
   released the texture Retain. Every presented frame leaked one IOSurface pin.
   Fix: `releaseAcquired()` called from `SurfaceTexture.Destroy()` and
   `DiscardTexture`.

2. **`nextDrawable` spiral** — with `allowsNextDrawableTimeout=true` (default),
   `nextDrawable` returned nil under pool pressure. The caller reconfigured the
   surface, allocating another pool — a positive-feedback leak loop.
   Fix: `allowsNextDrawableTimeout=false` (blocks until a drawable is free).

3. **CA stretch artifact** — `kCAGravityResize` (default) scaled the last
   rendered frame to fill the window on every AppKit resize tick.
   Fix: `contentsGravity="topLeft"`.

4. **Drawable stockpiling** — Core Animation holds the last-presented drawable
   for ~16 ms (one vsync) independently of `WaitIdle`. At 60+ Hz drag rate,
   old drawables accumulate faster than CA releases them.
   Fix: `presentsWithTransaction=true` during drag; drawable swap lands inside
   the open CA transaction — CA releases the old drawable immediately.

5. **Test crash** — `defer Release(desc)` on a factory-method return
   (`+0` autoreleased). LIFO ordering fired Release before pool.Drain, freeing
   the descriptor while child objects were still in the pool.
   Fix: removed the incorrect `defer Release`.

**UMA improvements (Apple Silicon):**

- `CreateTexture` uses `MTLStorageModeShared` on M-series; `WriteTexture`
  bypasses staging buffer with `replaceRegion:` for zero-copy CPU→GPU upload.
- `DestroyTexture` calls `setPurgeableState: empty` for immediate page reclaim.

**Frame throttling:** `frameSemaphore` (cap: 2 frames in-flight) prevents the
CPU from outrunning the GPU, bounding drawable pool occupancy to 3 drawables
(`setMaximumDrawableCount:3` matches Rust wgpu).

**Testing:** `go test ./hal/metal/... -count=5` passes consistently.


kolkov (approved):
LGTM! Thorough fix — all 5 root causes verified against Rust wgpu-hal and rio terminal.

**Cross-reference validation (Rust wgpu-hal `surface.rs` + `raphamorim/rio` `metal.rs`):**

| Pattern | Rust wgpu | rio | This PR |
|---------|-----------|-----|---------|
| Retain drawable+texture on acquire | `to_owned()`, drop releases | via wgpu | `Retain` + `releaseAcquired()` ✅ |
| `allowsNextDrawableTimeout` | `false` (line 178) | `false` (line 92) | `false` ✅ |
| `maximumDrawableCount` | latency+1 (2–3) | 3 (explicit) | 3 ✅ |
| `presentsWithTransaction` | field, default false | false | toggle during resize ✅ |
| nil drawable → no reconfigure | `Ok(None)` / Timeout | via wgpu | blocks (timeout=false) ✅ |

**The retain leak was the critical bug** — Rust releases via Drop trait, Go needed explicit `releaseAcquired()` in both `Destroy()` and `DiscardTexture`. Without it, every presented frame pinned one IOSurface permanently.

**Novel additions** (not in Rust wgpu but technically sound):
- `setPurgeableState(empty)` — aggressive page reclaim on DestroyTexture
- `frameSemaphore` cap=2 — explicit CPU-side throttle (Rust relies on drawable pool)
- UMA `StorageModeShared` + `replaceRegion:` direct writes — zero-copy on Apple Silicon
- `contentsGravity=topLeft` — Apple best practice, avoids stretch during resize

All verified against Apple Metal docs. `maintainAfterIdle()` correctly processes DestroyQueue entries after WaitIdle — matches Rust `LifetimeTracker::triage_submissions` pattern.

Test crash fix (removed `defer Release(desc)` on +0 autoreleased object) — correct ObjC memory management.

PR #252, opened 2026-07-12
More correction of GPU classification as integrated/discrete type for DX12

Zeroes1:
DedicatedVideoMemory — the amount of memory in bytes physically soldered on the GPU and accessible only by the GPU. For discrete GPUs this is real GDDR/HBM. For integrated Intel/AMD GPUs this is a reserve carved out of system DDR (typically 32–512 MB) that the driver "dedicates" and makes inaccessible to the CPU. The BIOS/UEFI allocates this region at boot.

SharedSystemMemory — the amount of system DDR that the GPU can use cooperatively with the CPU. This is a memory-mapped region accessible to both. For an iGPU this can be 8–32 GB (essentially all available RAM). For discrete GPUs this can also be > 0 (via PCIe BAR / Resizable BAR), but the bulk of video memory is still in DedicatedVideoMemory.

Key heuristic: if DedicatedVideoMemory is low (≤ 512 MB) and SharedSystemMemory is large, the GPU is guaranteed to be integrated — it has no dedicated VRAM and operates out of shared system DDR.

for other modes except DX12 - not required.

### DIFF
--- a/hal/dx12/adapter.go
+++ b/hal/dx12/adapter.go
@@ -281,22 +281,20 @@ func (a *Adapter) deviceType() gputypes.DeviceType {
 		return gputypes.DeviceTypeCPU
 	}
 
-	// Check for dedicated video memory to distinguish discrete from integrated
-	if a.desc.DedicatedVideoMemory > 0 {
-		// Heuristic: >512MB dedicated VRAM is likely discrete
-		if a.desc.DedicatedVideoMemory > 512*1024*1024 {
-			return gputypes.DeviceTypeDiscreteGPU
-		}
+	// Heuristic: >512MB dedicated VRAM is almost always a discrete GPU
+	// (NVIDIA/AMD with dedicated GDDR). Smaller amounts (≤512MB) are typically
+	// a reserved pool from system memory on integrated GPUs (Intel HD/UHD, AMD APU).
+	if a.desc.DedicatedVideoMemory > 512*1024*1024 {
+		return gputypes.DeviceTypeDiscreteGPU
 	}
 
-	// If there's no dedicated video memory, it's likely integrated
-	if a.desc.DedicatedVideoMemory == 0 && a.desc.SharedSystemMemory > 0 {
+	if a.desc.DedicatedVideoMemory > 0 {
 		return gputypes.DeviceTypeIntegratedGPU
 	}
 
-	// Assume discrete if there's any dedicated memory
-	if a.desc.DedicatedVideoMemory > 0 {
-		return gputypes.DeviceTypeDiscreteGPU
+	// No dedicated memory + shared system memory → integrated
+	if a.desc.SharedSystemMemory > 0 {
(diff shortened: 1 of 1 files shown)


kolkov (changes requested):
Thank you for the excellent analysis of `DedicatedVideoMemory` vs `SharedSystemMemory` semantics — the PR description is very well-researched!

The heuristic improvement is correct for common cases, but we found a more robust approach used by Rust wgpu that handles all edge cases without magic thresholds.

## Rust wgpu approach: UMA flag

Rust wgpu ([`wgpu-hal/src/dx12/adapter.rs:153-161`](https://github.com/gfx-rs/wgpu/blob/trunk/wgpu-hal/src/dx12/adapter.rs#L153-L161)) uses `CheckFeatureSupport(D3D12_FEATURE_ARCHITECTURE)` to query the `UMA` (Unified Memory Architecture) boolean directly from the driver:

```rust
device_type: if flags.contains(DXGI_ADAPTER_FLAG_SOFTWARE) {
    Cpu
} else if features_architecture.UMA.as_bool() {
    IntegratedGpu
} else {
    DiscreteGpu
}
```

This is the authoritative DX12 way — the hardware reports whether it has a dedicated memory bus or shares system RAM. No heuristics needed.

## Why UMA is better than memory thresholds

| Scenario | 512MB heuristic | UMA flag |
|----------|----------------|----------|
| Intel HD/UHD (iGPU, 128-512MB reserved) | ✅ Integrated | ✅ Integrated |
| AMD APU with >512MB BIOS reservation | ❌ **Discrete** (false positive) | ✅ Integrated |
| Intel Arc A770 (dGPU, dedicated GDDR6) | ✅ Discrete (usually >512MB) | ✅ Discrete |
| Snapdragon X Elite (ARM, Adreno iGPU) | ❓ Depends on reservation | ✅ Integrated |
| Old low-end dGPU with ≤512MB VRAM | ❌ **Integrated** (false positive) | ✅ Discrete |

## Good news: we already have most of the plumbing

- `D3D12_FEATURE_ARCHITECTURE = 1` constant — already defined in `d3d12/const.go`
- `CheckFeatureSupport` binding — already working (used for shader model queries)
- `probeCapabilities()` with temp device — already creates device and queries features

What's needed (~25 lines total):
1. Add `D3D12_FEATURE_DATA_ARCHITECTURE` struct to `d3d12/types.go` (4 fields, 16 bytes)
2. Add `IsUMA bool` to `AdapterCapabilities`
3. Add `queryArchitecture()` method (same pattern as existing `queryD3D12Options()`)
4. Simplify `deviceType()` to the 3-line Rust pattern

Would you like to update this PR with the UMA approach? We can guide you through the implementation — it follows the exact same pattern as the existing `queryD3D12Options()` in `adapter.go:135-159`. Happy to help with any questions!

Zeroes1:
yep, sure. This approach would be more correct.

PR #257, opened 2026-07-14
fix(vulkan): reject pipelines without layouts

besmpl:
## Summary

Reject nil Vulkan render and compute pipeline layouts before building native pipeline descriptors. The native HAL does not derive automatic layouts, so forwarding VK_NULL_HANDLE is invalid and can crash drivers rather than return a VkResult.

This extracts the independent Vulkan correctness fix from #253 and closes the same pre-FFI hazard for both pipeline types. It contains no prepared-indexed, MDI, Hearth, or module-path changes.

## Verification

- go test ./hal/vulkan
- GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go test -c -o /tmp/wgpu-vulkan-linux-amd64.test ./hal/vulkan
- git diff --check

### DIFF
--- a/hal/vulkan/pipeline.go
+++ b/hal/vulkan/pipeline.go
@@ -24,6 +24,15 @@ func (d *Device) CreateRenderPipeline(desc *hal.RenderPipelineDescriptor) (hal.R
 	if desc == nil {
 		return nil, fmt.Errorf("BUG: render pipeline descriptor is nil in Vulkan.CreateRenderPipeline — core validation gap")
 	}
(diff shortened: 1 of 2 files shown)


kolkov (approved):
Clean fix. Our core layer allows nil pipeline layout to reach the HAL (`pipeline.go:83` — skips validation when `Layout.IsZero()`), so this guard is necessary. Rust wgpu avoids this structurally with non-nullable types, but our HAL uses pointer interfaces.

Test approach is smart — zero-value Device proves the guard fires before any FFI.

@lkmavi — quick look welcome, this is a small defensive guard.

PR #258, opened 2026-07-14
fix(dx12): match indirect descriptor ABI size

besmpl:
## Summary

Match the native D3D12_INDIRECT_ARGUMENT_DESC layout: its largest union member is three uint32 values, so the union is 12 bytes and the full descriptor is 16 bytes. Add exact size and offset assertions.

This extracts the independent DX12 ABI correctness fix from #253. It contains no counted-draw or prepared-indexed behavior.

## Verification

- GOOS=windows GOARCH=amd64 CGO_ENABLED=0 go test -c -o /tmp/wgpu-d3d12-windows-amd64.test.exe ./hal/dx12/d3d12
- GOOS=windows GOARCH=arm64 CGO_ENABLED=0 go test -c -o /tmp/wgpu-d3d12-windows-arm64.test.exe ./hal/dx12/d3d12
- git diff --check

### DIFF
--- a/hal/dx12/d3d12/types.go
+++ b/hal/dx12/d3d12/types.go
@@ -597,7 +597,11 @@ type D3D12_COMMAND_SIGNATURE_DESC struct {
 type D3D12_INDIRECT_ARGUMENT_DESC struct {
 	Type D3D12_INDIRECT_ARGUMENT_TYPE
 	// Union for different argument types
-	Union [8]byte
+	// The largest member (D3D12_INDIRECT_ARGUMENT_DESC_CONSTANT) is three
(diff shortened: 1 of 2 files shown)


kolkov (approved):
Verified against Windows SDK `d3d12.h` — the largest union member is `D3D12_INDIRECT_ARGUMENT_DESC_CONSTANT` (three uint32 = 12 bytes), total struct = 16 bytes. Our old `[8]byte` was 4 bytes short.

The `unsafe.Sizeof` / `unsafe.Offsetof` assertions are exactly what we want for ABI correctness. Good find.

@lkmavi — FYI, small DX12 ABI fix. Worth a glance if you have a moment.

PR #259, opened 2026-07-14
fix(dx12): transition render targets explicitly

besmpl:
## Summary

Track DX12 texture resource state consistently and emit required barriers for:

- render-target and depth/stencil attachments
- buffer/texture and texture/texture copies
- MSAA resolve source/destination and resting states
- external surfaces around PRESENT

Textures created by this backend do not opt into ALLOW_SIMULTANEOUS_ACCESS, so render-target and depth states are not implicit promotions. Resolve transitions now use and update actual tracked states instead of hardcoded before-states. Focused tests capture exact transition sequences without calling native FFI.

This extracts and completes the independent DX12 transition fix identified in #253. It contains no counted-draw or prepared-indexed behavior.

## Verification

- GOOS=windows GOARCH=amd64 CGO_ENABLED=0 go test -c -o /tmp/wgpu-dx12-transitions-windows-amd64.test.exe ./hal/dx12
- GOOS=windows GOARCH=arm64 CGO_ENABLED=0 go test -c -o /tmp/wgpu-dx12-transitions-windows-arm64.test.exe ./hal/dx12
- git diff --check

### DIFF
--- a/hal/dx12/command.go
+++ b/hal/dx12/command.go
@@ -288,6 +288,7 @@ func (e *CommandEncoder) CopyBufferToTexture(src hal.Buffer, dst hal.Texture, re
 
 	// Transition source buffer to COPY_SOURCE if needed.
 	e.transitionBufferIfNeeded(srcBuf, d3d12.D3D12_RESOURCE_STATE_COPY_SOURCE)
+	e.transitionTextureIfNeeded(dstTex, d3d12.D3D12_RESOURCE_STATE_COPY_DEST)
 
 	for _, r := range regions {
 		// Source location (buffer)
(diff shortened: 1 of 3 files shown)


besmpl:
Marking this draft while I tighten the state model. Final review found that the current revision still relies on global whole-texture recording-time state, which is not safe for abandoned/reordered command buffers, swapchain PRESENT state, disjoint subresources, or sampled read-only depth. I will either replace it with command-local subresource-aware tracking and focused coverage, or narrow the PR to a correctness shape that can be proven independently. The current branch should not be merged.

besmpl:
Superseded by #262, which replaces the recording-time transition patch with submission-ordered state reconciliation and includes the reviewed copy/view/lifetime fixes. Please review #262 instead; this draft remains intentionally unmerged.

PR #260, opened 2026-07-14
fix(metal): pin autorelease pools to OS threads

besmpl:
## Summary

- pin each Objective-C autorelease pool to the OS thread where it is created
- make drain terminal and balance the thread lock even when creation or drain panics
- defer pool cleanup immediately at encoder call sites so recovered panics cannot strand a goroutine on an OS thread
- cover nesting, zero/double drain, panic cleanup, scheduler yields, and Metal resource stress

This is extracted from #253 as an independent Metal runtime correctness fix. It does not include indexed-indirect, ICB, pipeline, or material-page changes.

## Verification

- `CGO_ENABLED=0 go test -count=1 ./hal/metal`
- `CGO_ENABLED=1 go test -race -count=1 -run '^TestAutoreleasePool' ./hal/metal`
- `GOOS=darwin GOARCH=arm64 CGO_ENABLED=0 go test -run '^$' ./hal/metal`
- repeated native autorelease and resource stress tests on Apple Silicon
- `git diff --check`

The full Metal race suite still encounters a pre-existing goffi/checkptr failure in `TestCAMetalLayerDrawableSize`; the focused autorelease race suite passes.

### DIFF
--- a/hal/metal/encoder.go
+++ b/hal/metal/encoder.go
@@ -42,9 +42,9 @@ func (e *CommandEncoder) BeginEncoding(label string) error {
 	// This prevents LIFO violations when pools from different frames overlap
 	// on the ObjC autorelease pool stack (macOS Tahoe SIGABRT fix).
 	pool := NewAutoreleasePool()
+	defer pool.Drain()
 	e.cmdBuffer = MsgSend(e.device.commandQueue, Sel("commandBuffer"))
 	if e.cmdBuffer == 0 {
-		pool.Drain()
 		return fmt.Errorf("metal: failed to create command buffer")
 	}
 	Retain(e.cmdBuffer)
@@ -53,7 +53,6 @@ func (e *CommandEncoder) BeginEncoding(label string) error {
 		_ = MsgSend(e.cmdBuffer, Sel("setLabel:"), uintptr(nsLabel))
 		Release(nsLabel)
 	}
-	pool.Drain()
 	hal.Logger().Debug("metal: encoding started", "label", label)
 	return nil
 }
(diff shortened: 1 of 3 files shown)


kolkov (commented):
This addresses a real latent issue — we have 30+ autorelease pools across `device.go`, `encoder.go`, `queue.go`, `surface.go` without `runtime.LockOSThread`. Go 1.14+ async preemption can theoretically migrate a goroutine between pool create and drain, which would mean draining an ObjC pool on a wrong OS thread.

The implementation looks solid — panic-safe drain, double-drain no-op, constructor rollback, callback injection for testing.

We'd like @lkmavi to review and test this on macOS before we merge — he's our Metal/ObjC expert and has been actively testing the Metal backend. @lkmavi could you take a look at this PR and check if you've seen any intermittent issues that might be related to autorelease pool thread affinity?

Note: this touches `encoder.go` — PR #261 also touches it in different functions. This one should merge first if both are accepted.

lkmavi:
@kolkov I ran the examples, and no problems arose.

kolkov (approved):
Real Go-specific issue, well-engineered fix. Go 1.14+ async preemption can migrate goroutines between OS threads. ObjC autorelease pools are per-OS-thread (TLS-based stack). Without `LockOSThread`, a pool created on thread A could be drained on thread B after goroutine migration — LIFO violation or leak. Rust doesn't need this because Rust threads are 1:1 with OS threads.

All autorelease pool callsites across Metal files are automatically covered since the fix is in `NewAutoreleasePool()` itself. The encoder.go defer conversion is strictly better than the manual drain patterns it replaces.

Test coverage is excellent: 9 tests including panic recovery paths and concurrent stress with `runtime.Gosched()` yields. @lkmavi confirmed examples pass on macOS.

One optional nit: in `Drain()`, the `locked` field serves dual role as "thread is pinned" and "pool is alive" guard. A one-line comment clarifying this would help future readers. Not blocking.

LGTM.

PR #261, opened 2026-07-14
fix(metal): preserve texture array copy shape

besmpl:
## Summary

- represent WebGPU 1D/2D array textures with Metal `arrayLength` and physical depth 1
- copy array layers as Metal slices while preserving true 3D Z/depth semantics
- validate complete `Queue.WriteTexture` source ranges before any Metal effect, including compressed block geometry
- use dimension-correct Metal row/image strides and checked copy arithmetic

## Why

The previous Metal descriptor stored `DepthOrArrayLayers` in `depth` for every texture dimension, and copy paths used origin Z as both a texture slice and a region Z coordinate. That is correct for true 3D textures but not for 1D/2D arrays. It can create the wrong physical texture shape and copy only or address the wrong array layer. This also blocked a portable standard texture-array page in downstream renderers.

## Scope

This is an independent Metal correctness fix. It contains no multi-draw, prepared-command, ICB, material-page, or Hearth API surface.

## Verification

- `go test -count=1 ./hal/metal`
- focused physical copy/upload matrix repeated 20 times
- `CGO_ENABLED=0 go test ./...`
- `git diff --check HEAD^..HEAD`

Physical tests cover 1D arrays, 2D arrays, true 3D, both mixed 3D↔2D-array copy directions, private and shared uploads, non-zero origins/offsets, padded rows/images, and truncated final layers. Full race/vet remain blocked by pre-existing Objective-C/goffi unsafe/checkptr diagnostics outside this diff.

### DIFF
--- a/hal/metal/device.go
+++ b/hal/metal/device.go
@@ -190,11 +190,11 @@ func (d *Device) CreateTexture(desc *hal.TextureDescriptor) (hal.Texture, error)
 	_ = MsgSend(texDesc, Sel("setWidth:"), uintptr(desc.Size.Width))
 	_ = MsgSend(texDesc, Sel("setHeight:"), uintptr(desc.Size.Height))
 
-	depth := desc.Size.DepthOrArrayLayers
-	if depth == 0 {
-		depth = 1
+	shape := metalTextureDescriptorShape(desc.Dimension, desc.Size.DepthOrArrayLayers)
+	_ = MsgSend(texDesc, Sel("setDepth:"), uintptr(shape.depth))
+	if texType == MTLTextureType1DArray || texType == MTLTextureType2DArray {
+		_ = MsgSend(texDesc, Sel("setArrayLength:"), uintptr(shape.arrayLength))
 	}
-	_ = MsgSend(texDesc, Sel("setDepth:"), uintptr(depth))
 
 	mipLevels := desc.MipLevelCount
 	if mipLevels == 0 {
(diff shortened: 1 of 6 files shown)


kolkov (commented):
This fixes a real Metal API misuse — setting `depth` for all texture dimensions instead of `arrayLength` for 2D arrays. Matches Rust wgpu `device.rs:436-453`.

The copy decomposition (`metalCopyPlan`) and test coverage (648 lines) are thorough. However, our current consumers (gg, ui) only use `DepthOrArrayLayers: 1` — this bug doesn't affect anyone today.

@lkmavi — could you review this from the Metal side? You have the most hands-on experience with our Metal backend. We'd like your assessment before merging 1129 lines of Metal texture handling changes.

Depends on #260 merging first (both touch `encoder.go`).

kolkov (approved):
Real Metal API misuse fixed. The previous code set `setDepth:` for all texture dimensions instead of `setArrayLength:` for 1D/2D arrays. Matches Rust wgpu `device.rs:436-453` exactly.

The copy decomposition (`metalCopyPlan`) correctly handles all dimension combinations: 1D array, 2D array, 3D, and cross-dimension (3D↔2D-array). The upgrade to 6-argument `replaceRegion` (slice-aware form) and the `validateMetalTextureDataCopyLayout` with overflow protection are both correct.

Physical round-trip tests on real Metal hardware (buffer→texture→buffer, byte-by-byte verification) are the strongest evidence of correctness. Compressed format block info covers all BC/ETC2/EAC/ASTC formats.

No impact on existing consumers — gg/ui use `DepthOrArrayLayers=1`, which produces identical behavior to the old code.

Minor: `createMetalTextureCopyTestArray` is defined but never called — can remove. `Texture.depth` field stores `DepthOrArrayLayers` (not physical Metal depth) — a clarifying comment would help.

LGTM. Should merge after #260 (encoder.go dependency).

PR #264, opened 2026-07-15
fix(device): drain GPU work before teardown

besmpl:
## Summary

Make `Device.Release` wait for the HAL device to become idle before releasing pending writes, flushing deferred destruction, destroying pooled encoders, and finally destroying the core/HAL device.

The public `WaitIdle` method still rejects calls after release. Teardown uses a private `waitIdle` path because `Release` marks the device unavailable first, preventing concurrent new work while the final submission is drained.

## Why

The previous order set the released bit and then called the public `WaitIdle`, which immediately returned `ErrReleased`. Deferred encoder-recycling callbacks could therefore remain queued while HAL destruction reclaimed the same command pools.

## Validation

- `go test ./...`
- `go test -race .`
- a lifecycle test proves `WaitIdle` happens before HAL `Destroy`

This is intentionally separate from the Android work; it fixes native teardown ordering on every backend.

## Android integration relationship

This Android-independent change is one of five prerequisites replayed in [#268](https://github.com/gogpu/wgpu/pull/268). Its canonical candidate head is `6acfb1ea1cca1dc14d7ab0902931c678dc13e0a6`; the corresponding replay in #268 ends at `b2842f068d80d22eae297059aa983965965a9151`. The #268 description records the current exact integration head and full replay map; it drops this replay after this PR merges. Rust-v29 typed surface parity remains separate in [#273](https://github.com/gogpu/wgpu/pull/273). Android runtime and release claims remain in #268.

### DIFF
--- a/device_native.go
+++ b/device_native.go
@@ -863,6 +863,14 @@ func (d *Device) WaitIdle() error {
 	if d.released.Load() {
 		return ErrReleased
 	}
+	return d.waitIdle()
+}
+
+// waitIdle drains the HAL and deferred resource queues without consulting the
+// public released bit. Release marks the device unavailable before starting
+// teardown, but the native device must remain alive while its last submission
+// is drained.
+func (d *Device) waitIdle() error {
 	halDevice := d.halDevice()
 	if halDevice == nil {
 		return ErrReleased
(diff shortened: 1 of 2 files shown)


besmpl:
Stacked consumer preview: #268 at exact head `b11d2671e80081a191705c41c4e77e8b61472017` incorporates this PR's head `0ed17064f8c977f35d9b49b5cde0d0c69e867ecf` (matching stable patch ID).

The preview remains explicitly draft/non-mergeable and will be rebased to Android-only commits after this prerequisite lands.

kolkov (approved):
Good catch. The released flag was short-circuiting `WaitIdle()` during `Release()`, meaning the GPU was never actually drained before staging buffers and encoders were destroyed — a real use-after-free risk on all backends, though most drivers mask it with internal refcounting.

The private `waitIdle()` approach is clean: preserves the public API guard while allowing teardown to drain properly. The reorder of `queue.release()` to after `waitIdle()` matches Rust wgpu's `Queue::Drop` ordering where `pending_writes` is dropped after the fence wait completes (queue.rs:53, 157-251).

The lifecycle test with noop backend embedding is a solid way to assert the `[wait-idle, destroy]` invariant without requiring real GPU hardware.

LGTM.

PR #268, opened 2026-07-15
docs(android): reconcile preview integration after #273

besmpl:
## Status

Ready to merge at exact head `3a7acf9fa217db2ee2bfd3d7b5223de66f47df30`, directly on current WGPU `main` (`93fb27ff7ad717943e870580892b7490961a3ab7`). The PR is conflict-free and GitHub reports `MERGEABLE` / `CLEAN`. [CI run 30080220791](https://github.com/gogpu/wgpu/actions/runs/30080220791) is fully green, including both Android arm64 jobs.

## Rebase result

While resolving the requested rebase, I found that [#273](https://github.com/gogpu/wgpu/pull/273) had been squash-merged with the complete #268 Android stack still below its typed-surface commits. Its merge commit `93fb27ff7ad717943e870580892b7490961a3ab7` explicitly includes:

- `feat(vulkan): add Android arm64 WSI parity`;
- `ci(android): prove arm64 preview boundaries`;
- `docs(android): define the Vulkan preview contract`;
- `test(android): cover API 36 platform policy`; and
- `build(android): consume canonical goffi v0.6.1`.

I verified the result at the tree level: every file changed by #268's Android-owned range (`0eb4143..13b59db`) is byte-identical between the old #268 head and current `main`. Replaying those commits would therefore duplicate code that is already merged rather than produce a reviewable Android delta.

The clean rebase is consequently current `main` plus one focused documentation reconciliation commit. This preserves #268 as the original design/review record without manufacturing a duplicate implementation diff.

## What this PR now changes

- records that the Android Vulkan preview is already present on WGPU `main` through #273;
- removes obsolete prerequisite/rebase instructions;
- records canonical goffi v0.6.1 as the default backend dependency;
- keeps the optional Rust path pinned to the exact merged webgpu#24 source until a canonical release contains it; and
- keeps the API 29/API 36 physical-device matrix as an explicit release-evidence gate.

## Implementation review location

The Android implementation to review is now in [merge commit `93fb27f`](https://github.com/gogpu/wgpu/commit/93fb27ff7ad717943e870580892b7490961a3ab7), with the design, ownership, API-floor, and evidence discussion retained in this PR. The relevant implementation remains in `hal/vulkan/*android*`, `hal/vulkan/swapchain*`, `hal/allbackends/register_android.go`, and `scripts/check-android-arm64-preview.sh`.

## Verification

All current GitHub checks pass:

- Android arm64 preview with Go 1.25.12 and Go 1.26.5;
[...]

PR #275, opened 2026-07-24
test(metal): align storage-mode fixture with GPU policy

besmpl:
## Summary

- drive the layered `Queue.WriteTexture` storage-mode matrix through `Device.isAppleGPU`, the predicate production texture allocation actually uses
- skip the Shared-direct cases when the host does not expose an Apple GPU family
- leave Metal production behavior unchanged

## Root cause

The test added in #261 simulated Private and Shared texture allocation by changing `hasUnifiedMemory`. After #272, texture allocation intentionally uses Apple GPU-family membership (`isAppleGPU`) instead, avoiding Shared texture policy on Intel/AMD devices.

On Apple Silicon, the stale fixture therefore could not force the `private staging` case: `CreateTexture` still saw `isAppleGPU=true`, allocated Shared storage, and failed the expectation:

```text
texture shared mode = true; want false
```

This updates the test injection point to match the current policy. The Private-staging and Shared-direct data paths remain exercised on Apple GPUs; non-Apple GPUs continue to exercise the production-supported Private path.

## Validation

- `CGO_ENABLED=0 GOWORK=off go test -count=1 ./hal/metal -run '^TestQueueWriteTextureLayeredShapes/1D_array/private_staging$' -v`
- full layered storage matrix, repeated three times
- `CGO_ENABLED=0 GOWORK=off go test -count=1 ./hal/metal`
- `CGO_ENABLED=0 GOWORK=off go test -count=1 ./...`
- `CGO_ENABLED=0 GOWORK=off go test -count=1 -tags rust ./...`
- default and Rust-tagged builds, plus the root WASM build

The complete local macOS default and Rust-tagged test runs now pass without excluding `hal/metal`. This is deliberately separate from Android PR #268.

### DIFF
--- a/hal/metal/texture_copy_test.go
+++ b/hal/metal/texture_copy_test.go
@@ -450,7 +450,7 @@ func TestCommandEncoderCopiesBetweenLayeredTextures(t *testing.T) {
 
 func TestQueueWriteTextureLayeredShapes(t *testing.T) {
 	device, queue := newMetalTextureCopyTestDevice(t)
-	hasUnifiedMemory := device.hasUnifiedMemory
+	supportsSharedTextures := device.isAppleGPU
 
 	shapes := []struct {
 		name      string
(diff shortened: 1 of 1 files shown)


kolkov (commented):
Clean fix. The test injection was stale after #272 switched texture storage policy from `hasUnifiedMemory` to `isAppleGPU` — this aligns the fixture with the actual production predicate. Verified that `CreateTexture` (device.go:233) uses `d.isAppleGPU`, so the old `hasUnifiedMemory` toggle was a no-op for texture allocation. LGTM.

PR #276, opened 2026-07-24
feat(software): expose headless surface readback

besmpl:
Closes #256.

I’m very excited to get this foundation upstream: it makes deterministic, display-free software rendering available through the normal public surface lifecycle, while keeping backend ownership explicit and the API small.

## What this adds

- `wgpu.HeadlessSurfaceTarget`, a zero-sized safe target backed by the internal `hal.SurfaceTargetHeadless` delivered in #273.
- `Surface.ReadPixels()`, returning an owned, tightly packed, top-left RGBA8 snapshot after present/discard.
- An optional `hal.PixelReader` capability rather than widening the mandatory `hal.Surface` interface.
- Software readback for RGBA/BGRA, including sRGB variants and normalization to RGBA.
- Public clear and WGSL-triangle lifecycle proofs, ownership/error-state coverage, and a complete runnable documentation example.
- `software.Surface.GetFramebuffer()` remains as a compatibility alias.

This is intentionally a **Go software extension**, not a claim that Rust `wgpu` exposes a headless `SurfaceTarget`. Rust-tag and browser builds retain a compatible Go method set but reject the target/readback explicitly. Other native backends remain unchanged and unsupported; ordinary WebGPU GPU readback continues to use texture-to-buffer copies.

## Public flow

```go
surface, err := instance.CreateSurfaceFromTarget(wgpu.HeadlessSurfaceTarget{})
adapter, err := instance.RequestAdapter(&wgpu.RequestAdapterOptions{
    CompatibleSurface:    surface,
    ForceFallbackAdapter: true,
})
device, err := adapter.RequestDevice(nil)
err = surface.Configure(device, &wgpu.SurfaceConfiguration{ /* extent + format */ })

// Acquire -> render -> submit -> present through the ordinary surface API.
pixels, err := surface.ReadPixels() // owned RGBA8 snapshot
```

`docs/SURFACE-TARGETS.md` contains the complete compilable acquire/render/submit/present/readback example.

## Verification

Passed locally with `CGO_ENABLED=0` and `GOWORK=off`:

- `go test -count=1 ./hal/software ./core .`
- every repository package except `./hal/metal`
- `golangci-lint run --timeout=5m --new-from-rev=upstream/main` (`0 issues`)
- Rust-tag API compile
- `GOOS=js GOARCH=wasm` API compile
- Linux amd64 and Windows amd64 full cross-compilation
- documented public example (`64` output bytes)
- `git diff --check`

[...]

PR #277, opened 2026-07-24
fix(surface): handle nil WritePixels receiver

besmpl:
## Summary

- return `ErrReleased` when `WritePixels` is called on a nil `*Surface`
- match the nil-receiver behavior added for `ReadPixels` in #276
- add a focused regression proving the call returns an error instead of panicking

This addresses the sole non-blocking follow-up noted during review of #276. It does not change the behavior of configured, released, or unsupported surfaces.

## Validation

Passed locally with `CGO_ENABLED=0` and `GOWORK=off`:

- `go test -count=1 . -run '^TestSurfaceWritePixelsNil$'`
- `go test -count=1 ./hal/software ./core .`
- `go test -count=1 ./...`
- `go test -race -count=1 .`
- `go test -count=1 -tags rust ./...`
- default and Rust-tagged builds
- browser WASM root build
- Android arm64 full build
- `golangci-lint v2.12.2 run --timeout=5m --new-from-rev=upstream/main ./...`

Follow-up to #276.

### DIFF
--- a/headless_surface_native_test.go
+++ b/headless_surface_native_test.go
@@ -363,3 +363,10 @@ func TestHeadlessSurfaceReadPixelsStateErrors(t *testing.T) {
 		}
 	})
 }
+
+func TestSurfaceWritePixelsNil(t *testing.T) {
+	var surface *Surface
+	if err := surface.WritePixels(nil, 0, 0); !errors.Is(err, ErrReleased) {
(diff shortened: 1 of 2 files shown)


kolkov (commented):
Exactly the follow-up from our #276 review. `s == nil` guard now matches `ReadPixels`. Test covers the nil-receiver panic path. LGTM.

PR #306, opened 2026-08-09
fix(metal): bind buffer sizes for runtime arrays

### LINKED ISSUE #305: Metal: arrayLength() returns an incorrect value for runtime-sized storage arrays
- **Repo:** `github.com/gogpu/wgpu`
- **Version:** `v0.30.37` (with `github.com/gogpu/naga v0.18.0`)
- **Platform:** macOS 26.6.1 (Darwin 25.6.0), Apple M4, Metal backend, Go 1.25

## Summary

On the macOS Metal backend, `arrayLength()` returns an incorrect value for a runtime-sized storage array.

For example, when 16 bytes (4 x `u32`) are bound, the shader writes `arrayLength(&out)` and reads `1073741824` (`0x40000000`), not `4`.

Older releases have the same gap in the code, so this does not look like a recent regression.

**I plan to open a PR with a fix.** I am documenting the Metal-specific problem here so others can find it more easily.

## Reproduction

Save the following `main.go` in an empty module on macOS (Apple Silicon) and run it. The build steps and exit codes are in its doc comment.

<details><summary><code>main.go</code></summary>

```go
// Command arraylength-repro checks the WGSL contract that arrayLength() on a
// runtime-sized storage array returns the number of elements bound to the
// binding.
//
// It binds 16 bytes (4 x u32) and asks the shader for arrayLength(&out). The
// only correct answer is 4. One invocation writes it, so the result does not
// depend on dispatch size, scheduling, or thread index.
//
// Before it runs the shader, it prints the MSL that naga generates. Look at the
// _buffer_sizes argument there. It has no [[buffer(N)]] attribute.
//
// Run it with:
//
//	go mod init arraylength-repro
//	go get github.com/gogpu/wgpu@v0.30.37
//	go mod tidy
//	go run .
//
// The go mod tidy step matters. It resolves naga to the version that wgpu
// requires, so the printed MSL is the same MSL that wgpu uses.
//
// Exit status:
//
//	0  arrayLength() returned the correct value
//	1  arrayLength() returned the wrong value
//	2  the check could not be run: no usable GPU, or a wgpu call failed
//
// The program is a one-shot that exits as soon as it has an answer, so it does
// not bother releasing resources.
package main

import (
	"context"
	"encoding/binary"
	"fmt"
	"os"

	"github.com/gogpu/gputypes"
	"github.com/gogpu/naga"
	"github.com/gogpu/naga/msl"
	"github.com/gogpu/wgpu"

	// Register the platform's real HAL backends. Without this only the noop and
	// software backends are available, and neither executes compute.
	_ "github.com/gogpu/wgpu/hal/allbackends"
)

const shader = `
@group(0) @binding(0) var<storage, read_write> out: array<u32>;

@compute @workgroup_size(1)
fn main() {
    out[0] = arrayLength(&out);
}
`

// elems is the number of u32 elements bound to the storage binding, and nbytes
// is the size of that binding. arrayLength(&out) must return elems.
const (
	elems  = 4
	nbytes = uint64(elems) * 4
)

// check aborts with exit status 2 when a wgpu call fails, because a failed
// setup call means the arrayLength() check never ran. It is deliberately not
// exit status 1: that status is reserved for "the check ran and the value was
// wrong", so a caller can tell the bug apart from an unusable environment.
func check(op string, err error) {
	if err != nil {
		fmt.Printf("%s: %v\n", op, err)
		os.Exit(2)
	}
}

// dumpMSL prints the MSL that naga generates for the shader above.
//
// This repeats the call that wgpu makes in hal/metal/device.go: naga.Parse,
// then naga.LowerWithSource, then msl.Compile with msl.DefaultOptions(). wgpu
// does not change the result afterwards, so this is the same MSL that wgpu
// gives to Metal.
//
// It is a repeat, not a capture. wgpu keeps the MSL in a local variable, so a
// program cannot read the exact bytes that wgpu used.
func dumpMSL() {
	ast, err := naga.Parse(shader)
	check("naga.Parse", err)

	module, err := naga.LowerWithSource(ast, shader)
	check("naga.LowerWithSource", err)

	source, info, err := msl.Compile(module, msl.DefaultOptions())
	check("msl.Compile", err)

	fmt.Printf("RequiresSizesBuffer = %v\n", info.RequiresSizesBuffer)
	fmt.Printf("---- generated MSL ----\n%s\n-----------------------\n\n", source)
}

func main() {
	dumpMSL()

	instance, err := wgpu.CreateInstance(&wgpu.InstanceDescriptor{Backends: wgpu.BackendsPrimary})
	check("CreateInstance", err)

	adapter, err := instance.RequestAdapter(&wgpu.RequestAdapterOptions{
		PowerPreference: gputypes.PowerPreferenceHighPerformance,
	})
	check("RequestAdapter", err)

	device, err := adapter.RequestDevice(&wgpu.DeviceDescriptor{Label: "arraylength-repro"})
	check("RequestDevice", err)

	info := adapter.Info()
	fmt.Printf("backend : %v\n", info.Backend)
	fmt.Printf("adapter : %s\n", info.Name)

	module, err := device.CreateShaderModule(&wgpu.ShaderModuleDescriptor{WGSL: shader})
	check("CreateShaderModule", err)

	bgl, err := device.CreateBindGroupLayout(&wgpu.BindGroupLayoutDescriptor{
		Entries: []wgpu.BindGroupLayoutEntry{{
			Binding:    0,
			Visibility: wgpu.ShaderStageCompute,
			Buffer:     &gputypes.BufferBindingLayout{Type: gputypes.BufferBindingTypeStorage},
		}},
	})
	check("CreateBindGroupLayout", err)

	layout, err := device.CreatePipelineLayout(&wgpu.PipelineLayoutDescriptor{
		BindGroupLayouts: []*wgpu.BindGroupLayout{bgl},
	})
	check("CreatePipelineLayout", err)

	pipeline, err := device.CreateComputePipeline(&wgpu.ComputePipelineDescriptor{
		Layout:     layout,
		Module:     module,
		EntryPoint: "main",
	})
	check("CreateComputePipeline", err)

	storage, err := device.CreateBuffer(&wgpu.BufferDescriptor{
		Size:  nbytes,
		Usage: gputypes.BufferUsageStorage | gputypes.BufferUsageCopyDst | gputypes.BufferUsageCopySrc,
	})
	check("CreateBuffer(storage)", err)

	// Zero the buffer so a non-zero readback can only come from the dispatch.
	err = device.Queue().WriteBuffer(storage, 0, make([]byte, nbytes))
	check("Queue.WriteBuffer", err)

	bindGroup, err := device.CreateBindGroup(&wgpu.BindGroupDescriptor{
		Layout:  bgl,
		Entries: []wgpu.BindGroupEntry{{Binding: 0, Buffer: storage, Offset: 0, Size: nbytes}},
	})
	check("CreateBindGroup", err)

	staging, err := device.CreateBuffer(&wgpu.BufferDescriptor{
		Size:  nbytes,
		Usage: gputypes.BufferUsageMapRead | gputypes.BufferUsageCopyDst,
	})
	check("CreateBuffer(staging)", err)

	encoder, err := device.CreateCommandEncoder(&wgpu.CommandEncoderDescriptor{})
	check("CreateCommandEncoder", err)

	pass, err := encoder.BeginComputePass(&wgpu.ComputePassDescriptor{})
	check("BeginComputePass", err)

	pass.SetPipeline(pipeline)
	pass.SetBindGroup(0, bindGroup, nil)
	pass.Dispatch(1, 1, 1)
	check("ComputePass.End", pass.End())

	encoder.CopyBufferToBuffer(storage, 0, staging, 0, nbytes)

	cmd, err := encoder.Finish()
	check("CommandEncoder.Finish", err)

	_, err = device.Queue().Submit(cmd)
	check("Queue.Submit", err)

	check("Buffer.Map", staging.Map(context.Background(), wgpu.MapModeRead, 0, nbytes))

	rng, err := staging.MappedRange(0, nbytes)
	check("Buffer.MappedRange", err)

	got := binary.LittleEndian.Uint32(rng.Bytes())

	fmt.Printf("binding : %d bytes (%d x u32)\n", nbytes, elems)
	fmt.Printf("want    : arrayLength(&out) == %d\n", elems)
	fmt.Printf("got     : arrayLength(&out) == %d (0x%08X)\n", got, got)

	if got != elems {
		fmt.Println("result  : FAIL")
		os.Exit(1)
	}

	fmt.Println("result  : PASS")
}

```

</details>

### Output

This is the whole run.

```text
RequiresSizesBuffer = true
---- generated MSL ----
// language: metal2.1
#include <metal_stdlib>
#include <simd/simd.h>

using metal::uint;
struct DefaultConstructible {
    template<typename T>
    operator T() && {
        return T {};
    }
};

struct _mslBufferSizes {
    uint size0;
};

typedef uint type_1[1];

kernel void main_(
  device type_1& out [[buffer(0)]]
, constant _mslBufferSizes& _buffer_sizes
) {
    if (uint(0) < 1 + (_buffer_sizes.size0 - 0 - 4) / 4) {
        out[0] = 1 + (_buffer_sizes.size0 - 0 - 4) / 4;
    }
    return;
}

-----------------------

backend : Metal
adapter : Apple M4
binding : 16 bytes (4 x u32)
want    : arrayLength(&out) == 4
got     : arrayLength(&out) == 1073741824 (0x40000000)
result  : FAIL
exit status 1
```

The value is deterministic: five consecutive runs all produced `0x40000000`. A single invocation writes it, so the result does not depend on dispatch size, scheduling, or thread index.

## Root cause

The reproduction generates MSL with the same calls that `wgpu` makes: `naga.Parse`, `naga.LowerWithSource`, and `msl.Compile(module, msl.DefaultOptions())`. This is the [sequence used in `hal/metal/device.go`](https://github.com/gogpu/wgpu/blob/46c6b770d15214bc5707ef62d7589b40963c24d8/hal/metal/device.go#L534C3-L534C70), and `wgpu` does not modify the result afterwards. Because `wgpu` keeps the MSL in a local variable, the output above is regenerated rather than captured.

The output points to two problems.

1. **`_buffer_sizes` is missing its `[[buffer(N)]]` attribute.**

    The kernel signature contains `, constant _mslBufferSizes& _buffer_sizes` with no attribute, while the `out` argument above it has `[[buffer(0)]]`.

    MSL has no `arrayLength()`, so naga computes the length from the extra entry-point argument `constant _mslBufferSizes&`.

    In `github.com/gogpu/naga@v0.18.0`, [`resolveBufferSizesBinding`](https://github.com/gogpu/naga/blob/v0.18.0/msl/internal/codegen/writer.go#L1301) returns `[[buffer(N)]]` only when `Options.PerEntryPointMap[entryPoint].SizesBuffer` is set, or `[[user(fake0)]]` when `FakeMissingBindings` is true. [`writeEntryPointParam`](https://github.com/gogpu/naga/blob/v0.18.0/msl/internal/codegen/functions.go#L762-L764) writes the returned attribute after the argument.

    [`msl.DefaultOptions()`](https://github.com/gogpu/naga/blob/v0.18.0/msl/msl.go#L344-L351) sets neither option: `PerEntryPointMap` is nil and `FakeMissingBindings` is false. The argument is therefore emitted without an attribute, as shown in the dump above.

2. **`wgpu` does not bind the size data.**

    `msl.Compile` returns [`TranslationInfo.RequiresSizesBuffer`](https://github.com/gogpu/naga/blob/v0.18.0/msl/msl.go#L328-L330), which is `true` for this shader and appears above the MSL in the reproduction output.

    However, `RequiresSizesBuffer`, `SizesBuffer` and `PerEntryPointMap` do not appear anywhere in `gogpu/wgpu`, so the required size data is never bound.

The measured value is consistent with `_buffer_sizes.size0 == 0`. In `uint`, `1 + (0 - 0 - 4) / 4` underflows: `0 - 0 - 4` becomes `0xFFFFFFFC`, division by `4` gives `0x3FFFFFFF`, and adding `1` gives the observed `0x40000000`. With the correct value, `size0 = 16`, the same expression gives `4`.

To check whether `_buffer_sizes` might be reading the storage buffer at `[[buffer(0)]]`, I filled that buffer with `0xAAAAAAAA` and ran the program again. That value would produce `0x2AAAAAAA` in the same expression, but the result remained `0x40000000`. This rules out the storage buffer as the source of `_buffer_sizes`.

I did not determine which buffer index Metal assigns to an argument with no attribute.

The guard in the generated MSL is naga's bounds check, not one written by the shader author. Under the default [`BoundsCheckReadZeroSkipWrite`](https://github.com/gogpu/naga/blob/v0.18.0/msl/msl.go#L334-L341) policy, it should skip an out-of-range write. The same underflow makes the guard pass instead, so the problem can also affect shaders that declare a runtime-sized array without calling `arrayLength()`.

The reproduction does not attempt an out-of-bounds write, so this report only identifies the mechanism; it does not claim that one occurred.

For comparison, Rust `wgpu` binds this argument. Its Metal backend [reserves a slot](https://github.com/gfx-rs/wgpu/blob/wgpu-v29.0.1/wgpu-hal/src/metal/device.rs#L851-L856) after the declared buffers, and passes the slot to naga as `EntryPointResources.sizes_buffer`. Then it [uploads the sizes](https://github.com/gfx-rs/wgpu/blob/wgpu-v29.0.1/wgpu-hal/src/metal/command.rs#L1736-L1747) with `setBytes:length:atIndex:` when it sets the compute pipeline.

## Why this looks Metal-specific

Only naga's MSL backend uses this extra argument.

The other backends in this repository read the size from the bound resource. They need no extra argument, so no argument can stay unbound.

- SPIR-V uses `OpArrayLength` ([`emitArrayLength`](https://github.com/gogpu/naga/blob/v0.18.0/spirv/internal/codegen/backend.go#L9583)).
- HLSL writes a [`NagaBufferLength`](https://github.com/gogpu/naga/blob/v0.18.0/hlsl/internal/codegen/writer.go#L994) helper that calls `ByteAddressBuffer.GetDimensions`.
- Direct DXIL uses `dx.op.getDimensions` on the resource handle ([`emitArrayLength`](https://github.com/gogpu/naga/blob/v0.18.0/dxil/internal/emit/expressions.go#L4202)). This is the second DX12 path in this repository. The other one is [`compileWGSLModuleHLSL`](https://github.com/gogpu/wgpu/blob/46c6b770d15214bc5707ef62d7589b40963c24d8/hal/dx12/device.go#L2207).
- GLSL writes [`uint(<array>.length())`](https://github.com/gogpu/naga/blob/v0.18.0/glsl/internal/codegen/expressions.go#L1914-L1920).

> [!NOTE]
> **What I did not measure:** I tested Metal only. I have no Vulkan, DX12 or GLES adapter on my machine. I read the four items above in the naga source. They show that this mechanism does not exist in those backends. They are not runtime tests of those backends.

## Workaround

**Do not use `arrayLength()` in WGSL that runs on Metal.**

Pass the element count as data instead. A uniform works, or a header field in a `var<storage>` binding. Then the shader does not use the wrong `arrayLength()` value.

However, this does not fix everything. If the shader still declares a runtime-sized array, naga still emits `_mslBufferSizes`, and still uses it for the implicit bounds checks. Those checks stay affected.

To avoid that, use a fixed-size array type for the binding and pass the actual count separately.

A pipeline override constant is not a workaround on this backend. [`hal/metal/device.go:886`](https://github.com/gogpu/wgpu/blob/46c6b770d15214bc5707ef62d7589b40963c24d8/hal/metal/device.go#L886-L891) says that `desc.Compute.Constants` is not applied yet.

## A possible fix

The fix in the upcoming PR follows Rust `wgpu`.

**Which buffer index to use.** Rust takes the index right after the declared buffers, and it knows that number while it builds the pipeline layout. This repository already computes the same number: [`CreatePipelineLayout`](https://github.com/gogpu/wgpu/blob/46c6b770d15214bc5707ef62d7589b40963c24d8/hal/metal/device.go#L485-L501) accumulates `bufAccum` over the bind group layouts. So the slot can come from the layout, and no guess is needed at encode time.

**Where to compile the MSL.** `PerEntryPointMap` is keyed by entry-point name, so the compile needs both the entry point and the layout. Rust compiles inside `create_compute_pipeline`, and its [`load_shader`](https://github.com/gfx-rs/wgpu/blob/wgpu-v29.0.1/wgpu-hal/src/metal/device.rs#L138-L145) takes the layout as an argument. This repository compiles earlier, in `CreateShaderModule`, where neither is known yet. So the patch moves that work to pipeline creation.

If you would rather keep the compile in `CreateShaderModule` and pass the slot some other way, tell me and I will redo it.

Two more things about the fix:

- It covers the compute path only. The render path emits the same argument, and I expect the same gap, but I have not reproduced it. It keeps compiling with `msl.DefaultOptions()`, so its MSL does not change. I would rather file that separately once I have a reproduction.
- It adds tests at two levels. One runs on a GPU and is verified locally, because the standard runners have no GPU. The others read the generated MSL, which is plain text, so they need no GPU and run in CI. They check that `_buffer_sizes` gets a `[[buffer(N)]]`, and that every other binding keeps the slot it had before. Without them, a regression here would be invisible again.

KEINOS:
Fixes #305

## Summary

On Metal, `arrayLength()` returns an incorrect value for a runtime-sized storage array. With 16 bytes (4 x `u32`) bound, the shader reads `1073741824` (`0x40000000`) instead of `4`.

naga's MSL backend reads the length from an extra entry point argument, `constant _mslBufferSizes& _buffer_sizes`. Before this PR, `hal/metal` did not bind that argument. This PR assigns it a buffer slot and uploads the buffer sizes before each compute dispatch, following the Rust Metal HAL.

## Root cause

MSL has no `arrayLength()`, so naga computes it as `1 + (size0 - offset - stride) / stride`, where `size0` comes from the `_buffer_sizes` argument. naga writes a `[[buffer(N)]]` attribute on that argument only when `msl.Options.PerEntryPointMap[ep].SizesBuffer` is set (`resolveBufferSizesBinding`, naga v0.18.0 `msl/internal/codegen/writer.go:1301`).

`CreateShaderModule` calls `msl.Compile(irModule, msl.DefaultOptions())` (`hal/metal/device.go:534`). `msl.DefaultOptions()` leaves `PerEntryPointMap` nil, so the argument is emitted with no attribute:

```metal
kernel void main_(
  device type_1& out [[buffer(0)]]
, constant _mslBufferSizes& _buffer_sizes
) {
    if (uint(0) < 1 + (_buffer_sizes.size0 - 0 - 4) / 4) {
        out[0] = 1 + (_buffer_sizes.size0 - 0 - 4) / 4;
    }
```

`msl.Compile` does report the requirement — `TranslationInfo.RequiresSizesBuffer` is `true` — but before this PR, `RequiresSizesBuffer`, `SizesBuffer` and `PerEntryPointMap` did not appear anywhere in the repository, so the required size data was never bound. The observed value is consistent with `size0` reading as zero: `1 + (0 - 0 - 4) / 4` wraps around in `uint` to `0x40000000`.

The same argument feeds naga's implicit bounds checks, whose default policy is `BoundsCheckReadZeroSkipWrite`. The same wrap-around makes those guards pass, so the problem can also affect a shader that never calls `arrayLength()`, as long as it declares a runtime-sized array.

## Changes

- **`hal/metal/device.go`** — `CreatePipelineLayout` stores the accumulated buffer count as `PipelineLayout.totalBuffers`. That is the first free Metal buffer slot and therefore the slot for `_buffer_sizes`; the `bufAccum` loop already computed the value.

  `CreateShaderModule` now stops after `naga.Parse`, `naga.LowerWithSource` and workgroup size extraction, and keeps the IR module. `msl.Compile` and `newLibraryWithSource:` move to a new `Device.compileLibrary`, called from `CreateComputePipeline` and `CreateRenderPipeline`. The slot is only known there. WGSL syntax and validation errors are still reported from `CreateShaderModule`.

  `CreateComputePipeline` passes `bindSizes = true` and records the bindings and the slot on the pipeline. `CreateRenderPipeline` passes `false` and keeps `msl.DefaultOptions()`, so **the MSL it produces does not change** — see Scope.

  Pipeline creation now fails instead of running a kernel with an unbound argument: when a shader uses a runtime-sized array but has no `*PipelineLayout`, and when naga reports `RequiresSizesBuffer` but no binding was found.

- **`hal/metal/mslmap`** (new package) — builds the `msl.Options`. The map has to list **every** resource, not only the sizes buffer: when an entry point is present in `PerEntryPointMap`, naga's `computeResourceMap` uses its `Resources` field as is and does no numbering of its own, so a map carrying only `SizesBuffer` would silently move every normal binding to a wrong slot. `mslmap.Options` therefore reproduces naga's numbering rule — globals that have a binding, sorted by group then binding, with samplers, textures and buffers counted apart from zero. That is the order `SetBindGroup` already binds in.

  The package has no build tags and does not call Objective-C, so its tests run on the Linux CI runner. `hal/metal` itself cannot host them, because `hal/metal/main_test.go` calls `os.Exit(0)` for the whole package when `CI` or `GITHUB_ACTIONS` is set.

- **`hal/metal/encoder.go`** — `SetBindGroup` records each bound buffer's usable byte size (the entry `Size`, or `MTLBuffer.length - offset` when `Size` is zero). `Dispatch` and `DispatchIndirect` upload those sizes with `setBytes:length:atIndex:` at the slot the pipeline was compiled with.

- **`hal/metal/resource.go`** — the new fields, and `ShaderModule` now holds the IR module instead of an `MTLLibrary`. The library is created and released per pipeline.

## Rust wgpu reference

`gfx-rs/wgpu` at `wgpu-v29.0.1`.

| Decision | Rust wgpu | Here |
| --- | --- | --- |
| Which slot | [`device.rs:851-856`](https://github.com/gfx-rs/wgpu/blob/wgpu-v29.0.1/wgpu-hal/src/metal/device.rs#L851-L856) — `info.sizes_buffer = Some(info.counters.buffers)`, fixed while the pipeline layout is built | `PipelineLayout.totalBuffers` |
| Where to compile | [`device.rs:138-145`](https://github.com/gfx-rs/wgpu/blob/wgpu-v29.0.1/wgpu-hal/src/metal/device.rs#L138-L145) — `load_shader` takes `layout: &super::PipelineLayout` | `Device.compileLibrary(module, layout, bindSizes)` |
| How to upload | [`command.rs:1736-1747`](https://github.com/gfx-rs/wgpu/blob/wgpu-v29.0.1/wgpu-hal/src/metal/command.rs#L1736-L1747) — `set_bytes` of the size array at that slot | `ComputePassEncoder.bindBufferSizes` |

The `GroupSlotOffsets` doc comment already cites `PipelineLayout.bind_group_infos` from the same Rust file, so the slot comes from the structure that was built for it.

## Testing

Apple M4, macOS 26.6.1, Go 1.25.

The GPU tests are in the **first commit**, so their failures are visible in the history. Before the fix, all three command sequences return the same incorrect value:

```text
--- FAIL: TestArrayLengthReportsBoundSize (0.03s)
--- FAIL: TestArrayLengthReportsBoundSizeWhenBindGroupPrecedesPipeline (0.00s)
--- FAIL: TestArrayLengthReportsBoundSizeAfterCompatiblePipelineSwitch (0.00s)

Each test reports:
arrayLength(&out) = 1073741824 (0x40000000); want 4
    backend Metal derived a runtime array length from an unbound buffer-sizes argument
```

After:

```text
$ go test -tags integration -run TestArrayLength -count=5 .
ok  	github.com/gogpu/wgpu	0.112s
```

It is behind `integration && darwin && !rust && !(js && wasm)`, so the default suite is unaffected.

The MSL tests need no GPU and run in CI:

```text
$ go test -v ./hal/metal/mslmap/
--- PASS: TestDefaultOptionsLeaveSizesBufferUnbound
--- PASS: TestOptionsBindSizesBuffer
--- PASS: TestOptionsPreserveOrdinaryBindings
--- PASS: TestOptionsRejectsOutOfRangeSlot
ok  	github.com/gogpu/wgpu/hal/metal/mslmap	0.005s
```

`TestOptionsPreserveOrdinaryBindings` is the one that guards the "complete map" requirement: it compiles the same shader with `msl.DefaultOptions` and with the explicit map, and requires every normal parameter to keep the same `[[buffer(N)]]`.

Examples on Metal:

```text
$ GOGPU_GRAPHICS_API=metal go run ./examples/compute-sum/
1. Creating instance... OK
2. Requesting adapter... OK (Apple M4)
3. Creating device... OK
4. Input: 256 elements, CPU sum = 32896
5. Creating buffers... OK
6. Creating compute pipeline... OK
7. Dispatching compute... OK
8. Reading results... OK

CPU reference sum: 32896
GPU partial sum:   32896
PASS: GPU sum matches CPU reference
```

`compute-copy`, `compute-particles` and `triangle-headless` also pass on Metal.

Pre-submit checklist: `go fmt ./...`, `go build ./...`, `go test ./...` and `go test -cover ./...` are clean.

```text
$ GOOS=linux GOARCH=amd64 golangci-lint run --timeout=5m
0 issues.
```

Four broader macOS checks still report pre-existing issues. I checked each result against `main` at `46c6b77`; this PR does not introduce them.

- `go vet ./...` reports four `possible misuse of unsafe.Pointer` findings, in `hal/software/blit_darwin.go` and `hal/metal/{objc,resource}.go`. Identical on `main`; only a line number moves, because this PR adds an import to `resource.go`.
- `go test -race ./...` fails in `hal/noop` (`TestNoopConcurrentAccess`). Identical on `main`. `go test -race ./hal/metal/... .` passes.
- `GOOS=darwin GOARCH=arm64 golangci-lint run` reports 11 issues on unchanged lines; `main` reports 13. The two removed findings were attached to code that this PR restructures. I left the remaining cleanup out of scope.
- `go test -tags integration ./...` has four failures and one crash in `TestTypedSurfaceCreationPathsUseBackendMap`. The set is identical on `main`.

Downstream, two projects of mine that dispatch compute kernels over runtime-sized arrays were hitting this bug and now pass their suites against this branch.

## Scope

- **Compute only.** The render path emits the same argument and I expect the same gap, but I have not reproduced it, so I did not change its behaviour: it still compiles with `msl.DefaultOptions()` and its MSL is byte for byte what it was. I would rather file that separately, with its own test.
- No changes to other backends, and no lint cleanup outside the bug.
- The `TODO(compute-constants)` and `TODO(zero-init-workgroup)` notes on `CreateComputePipeline` are untouched.

## Considered and rejected

**Tracking the slot at encode time.** My first version watched the highest buffer slot actually bound during `SetBindGroup` and put the sizes buffer after it. It works on the shaders I tested, but it is wrong in principle: a layout entry that no bind group fills, or a gap between groups, gives a different slot than the one the MSL was compiled against. Taking the slot from the pipeline layout is deterministic and is what Rust does.

**Keeping `msl.Compile` in `CreateShaderModule`.** This would need the slot to be passed in some other way, and `PerEntryPointMap` is keyed by entry point name, which is also unknown there. If you would rather have it that way and thread the values differently, say so and I will redo it — moving the code is cheap.

### DIFF
--- a/arraylength_repro_test.go
+++ b/arraylength_repro_test.go
@@ -1,0 +1,269 @@
+// Copyright 2025 The GoGPU Authors
+// SPDX-License-Identifier: MIT
+
+//go:build integration && darwin && !rust && !(js && wasm)
+
+package wgpu_test
+
+import (
+	"context"
+	"encoding/binary"
+	"testing"
+
+	"github.com/gogpu/gputypes"
+	"github.com/gogpu/wgpu"
+
+	// Register the Metal HAL backend. Without a real backend registered only
+	// the noop/software backends are available, and neither executes compute.
+	//
+	// This import is why the file sits behind the integration tag: registering
+	// a real backend changes adapter selection for every test in the package.
+	_ "github.com/gogpu/wgpu/hal/metal"
+)
+
+// arrayLengthShader writes arrayLength(&out) into the first element of a
+// runtime-sized storage array.
+//
+// WGSL defines arrayLength() as the number of elements bound to the binding, so
+// for a 16-byte binding of u32 the only correct answer is 4. One invocation
+// writes it, so the result cannot depend on dispatch size, scheduling, thread
+// index, or buffer placement.
+const arrayLengthShader = `
+@group(0) @binding(0) var<storage, read_write> out: array<u32>;
+
+@compute @workgroup_size(1)
+fn main() {
+    out[0] = arrayLength(&out);
+}
+`
+
+// noArrayLengthShader has no runtime-sized array, but uses the same explicit
+// pipeline layout as arrayLengthShader. It lets a test switch between compatible
+// pipelines without rebinding the bind group.
+const noArrayLengthShader = `
+@compute @workgroup_size(1)
+fn main() {}
+`
+
+// arrayLengthElems is the element count bound to the storage binding.
+const arrayLengthElems = 4
+
+type arrayLengthBindOrder uint8
+
+const (
+	pipelineBeforeBindGroup arrayLengthBindOrder = iota
+	bindGroupBeforePipeline
+	compatiblePipelineSwitch
+)
+
+// TestArrayLengthReportsBoundSize asserts the WGSL contract that arrayLength()
+// equals the number of elements bound to the binding.
+//
+// This currently FAILS on Metal, returning 0x40000000 instead of 4.
+//
+// naga's MSL backend cannot express arrayLength() directly — MSL has no
+// equivalent construct — so it lowers it to
+// `1 + (_buffer_sizes.sizeN - offset - elemSize) / stride`, reading an extra
+// `constant _mslBufferSizes&` entry-point argument. That argument only carries
+// a [[buffer(N)]] attribute when msl.Options.PerEntryPointMap[ep].SizesBuffer
+// is set, and CreateShaderModule passes msl.DefaultOptions(), which leaves the
+// map nil. naga still reports the requirement via
+// TranslationInfo.RequiresSizesBuffer, but hal/metal never reads that flag and
+// never creates or binds a sizes buffer.
+//
+// The observed 0x40000000 follows from _buffer_sizes.size0 reading as zero:
+// in uint, 1 + (0 - 0 - 4) / 4 underflows to 1 + 0x3FFFFFFF. Which buffer
+// index Metal assigns to the un-attributed argument is not verified here.
+//
+// The defect is specific to Metal by construction: this extra argument exists
+// only in naga's MSL backend. SPIR-V emits OpArrayLength, HLSL emits a
+// NagaBufferLength helper over ByteAddressBuffer.GetDimensions, DXIL emits
+// dx.op.getDimensions, and GLSL emits .length() — all derived from the bound
+// resource, so there is nothing extra left to bind.
+func TestArrayLengthReportsBoundSize(t *testing.T) {
+	testArrayLengthReportsBoundSize(t, pipelineBeforeBindGroup)
+}
+
+// TestArrayLengthReportsBoundSizeWhenBindGroupPrecedesPipeline verifies that
+// buffer-size state does not depend on whether the pipeline is set first.
+func TestArrayLengthReportsBoundSizeWhenBindGroupPrecedesPipeline(t *testing.T) {
+	testArrayLengthReportsBoundSize(t, bindGroupBeforePipeline)
+}
+
+// TestArrayLengthReportsBoundSizeAfterCompatiblePipelineSwitch verifies that a
+// compatible bind group remains usable when switching from a pipeline that
+// needs no sizes buffer to one that does.
+func TestArrayLengthReportsBoundSizeAfterCompatiblePipelineSwitch(t *testing.T) {
+	testArrayLengthReportsBoundSize(t, compatiblePipelineSwitch)
+}
+
+func testArrayLengthReportsBoundSize(t *testing.T, order arrayLengthBindOrder) {
+	t.Helper()
+
+	const u32Size = 4
+	const nbytes = uint64(arrayLengthElems) * u32Size
+
+	instance, err := wgpu.CreateInstance(&wgpu.InstanceDescriptor{Backends: wgpu.BackendsPrimary})
+	if err != nil {
+		t.Skipf("CreateInstance: %v", err)
+	}
+	defer instance.Release()
+
+	adapter, err := instance.RequestAdapter(&wgpu.RequestAdapterOptions{
+		PowerPreference: gputypes.PowerPreferenceHighPerformance,
+	})
+	if err != nil {
+		t.Skipf("RequestAdapter: %v", err)
+	}
+	defer adapter.Release()
+
+	device, err := adapter.RequestDevice(&wgpu.DeviceDescriptor{Label: "arraylength-repro"})
+	if err != nil {
+		t.Skipf("RequestDevice: %v", err)
+	}
+	defer device.Release()
+
+	info := adapter.Info()
+	t.Logf("adapter: backend=%v name=%q", info.Backend, info.Name)
+
+	shader, err := device.CreateShaderModule(&wgpu.ShaderModuleDescriptor{WGSL: arrayLengthShader})
+	if err != nil {
+		t.Fatalf("CreateShaderModule: %v", err)
+	}
+	defer shader.Release()
+
+	bgl, err := device.CreateBindGroupLayout(&wgpu.BindGroupLayoutDescriptor{
+		Entries: []wgpu.BindGroupLayoutEntry{{
+			Binding:    0,
+			Visibility: wgpu.ShaderStageCompute,
+			Buffer:     &gputypes.BufferBindingLayout{Type: gputypes.BufferBindingTypeStorage},
+		}},
+	})
+	if err != nil {
+		t.Fatalf("CreateBindGroupLayout: %v", err)
+	}
+	defer bgl.Release()
+
+	pipelineLayout, err := device.CreatePipelineLayout(&wgpu.PipelineLayoutDescriptor{
+		BindGroupLayouts: []*wgpu.BindGroupLayout{bgl},
+	})
+	if err != nil {
+		t.Fatalf("CreatePipelineLayout: %v", err)
+	}
+	defer pipelineLayout.Release()
+
+	pipeline, err := device.CreateComputePipeline(&wgpu.ComputePipelineDescriptor{
+		Layout:     pipelineLayout,
+		Module:     shader,
+		EntryPoint: "main",
+	})
+	if err != nil {
+		t.Fatalf("CreateComputePipeline: %v", err)
+	}
+	defer pipeline.Release()
+
+	var noSizesPipeline *wgpu.ComputePipeline
+	if order == compatiblePipelineSwitch {
+		noSizesShader, shaderErr := device.CreateShaderModule(&wgpu.ShaderModuleDescriptor{WGSL: noArrayLengthShader})
+		if shaderErr != nil {
+			t.Fatalf("CreateShaderModule(no arrayLength): %v", shaderErr)
+		}
+		defer noSizesShader.Release()
+
+		noSizesPipeline, err = device.CreateComputePipeline(&wgpu.ComputePipelineDescriptor{
+			Layout:     pipelineLayout,
+			Module:     noSizesShader,
+			EntryPoint: "main",
+		})
+		if err != nil {
+			t.Fatalf("CreateComputePipeline(no arrayLength): %v", err)
+		}
+		defer noSizesPipeline.Release()
+	}
+
+	storage, err := device.CreateBuffer(&wgpu.BufferDescriptor{
+		Label: "arraylength-storage",
+		Size:  nbytes,
+		Usage: gputypes.BufferUsageStorage | gputypes.BufferUsageCopyDst | gputypes.BufferUsageCopySrc,
+	})
+	if err != nil {
+		t.Fatalf("CreateBuffer(storage): %v", err)
+	}
+	defer storage.Release()
+
+	// Zero the buffer so a non-zero readback can only come from the dispatch.
+	device.Queue().WriteBuffer(storage, 0, make([]byte, nbytes))
+
+	bindGroup, err := device.CreateBindGroup(&wgpu.BindGroupDescriptor{
+		Layout:  bgl,
+		Entries: []wgpu.BindGroupEntry{{Binding: 0, Buffer: storage, Offset: 0, Size: nbytes}},
+	})
+	if err != nil {
+		t.Fatalf("CreateBindGroup: %v", err)
+	}
+	defer bindGroup.Release()
+
+	staging, err := device.CreateBuffer(&wgpu.BufferDescriptor{
+		Label: "arraylength-staging",
+		Size:  nbytes,
+		Usage: gputypes.BufferUsageMapRead | gputypes.BufferUsageCopyDst,
+	})
+	if err != nil {
+		t.Fatalf("CreateBuffer(staging): %v", err)
+	}
+	defer staging.Release()
+
+	encoder, err := device.CreateCommandEncoder(&wgpu.CommandEncoderDescriptor{})
+	if err != nil {
+		t.Fatalf("CreateCommandEncoder: %v", err)
+	}
+
+	pass, err := encoder.BeginComputePass(&wgpu.ComputePassDescriptor{})
+	if err != nil {
+		t.Fatalf("BeginComputePass: %v", err)
+	}
+	switch order {
+	case pipelineBeforeBindGroup:
+		pass.SetPipeline(pipeline)
+		pass.SetBindGroup(0, bindGroup, nil)
+	case bindGroupBeforePipeline:
+		pass.SetBindGroup(0, bindGroup, nil)
+		pass.SetPipeline(pipeline)
+	case compatiblePipelineSwitch:
+		pass.SetPipeline(noSizesPipeline)
+		pass.SetBindGroup(0, bindGroup, nil)
+		pass.SetPipeline(pipeline)
+	default:
+		t.Fatalf("unknown bind order %d", order)
+	}
+	pass.Dispatch(1, 1, 1)
+	if err := pass.End(); err != nil {
+		t.Fatalf("ComputePass.End: %v", err)
+	}
+
+	encoder.CopyBufferToBuffer(storage, 0, staging, 0, nbytes)
+
+	cmd, err := encoder.Finish()
+	if err != nil {
+		t.Fatalf("CommandEncoder.Finish: %v", err)
+	}
+	if _, err := device.Queue().Submit(cmd); err != nil {
+		t.Fatalf("Queue.Submit: %v", err)
+	}
+
+	if err := staging.Map(context.Background(), wgpu.MapModeRead, 0, nbytes); err != nil {
+		t.Fatalf("Buffer.Map: %v", err)
+	}
+	rng, err := staging.MappedRange(0, nbytes)
+	if err != nil {
+		t.Fatalf("Buffer.MappedRange: %v", err)
+	}
+	got := binary.LittleEndian.Uint32(rng.Bytes())
+	staging.Unmap()
+
+	if got != arrayLengthElems {
+		t.Errorf("arrayLength(&out) = %d (0x%08X); want %d\n"+
+			"backend %v derived a runtime array length from an unbound buffer-sizes argument",
+			got, got, arrayLengthElems, info.Backend)
+	}
+}
--- a/hal/metal/device.go
+++ b/hal/metal/device.go
@@ -18,6 +18,7 @@ import (
 	"github.com/gogpu/naga/ir"
 	"github.com/gogpu/naga/msl"
 	"github.com/gogpu/wgpu/hal"
+	"github.com/gogpu/wgpu/hal/metal/mslmap"
 )
 
 // Vertex buffer indices are assigned from the end of the range and count down.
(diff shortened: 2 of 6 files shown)


kolkov (approved):
Excellent work, @KEINOS. Verified every claim against Rust wgpu Metal HAL reference — architecture, slot computation, binding order, type detection all match. The `mslmap` package with cross-platform tests and `TestOptionsPreserveOrdinaryBindings` guard is particularly well done.

**Approving.** Two non-blocking notes for follow-up:

1. **Render pipeline `arrayLength()`** — you correctly scoped this to compute only (`bindSizes: false` for render). When we pick up the render side, `totalBuffers` will need to become per-stage (matching Rust's `per_stage_map` at `device.rs:832-839`). Will track as a follow-up issue.

2. **Minor perf** — `bindBufferSizes()` allocates `[]uint32` on every `Dispatch`. For hot compute paths (many dispatches per pass, same bindings), a pre-allocated slice on `ComputePassEncoder` would avoid this. Not blocking — correctness is fine.

KEINOS:
@kolkov 

Thank you for the follow-up note! I’d like to take the `bindBufferSizes()` allocation improvement and send it as a separate PR.

I’ll open the draft PR with the functional coverage test commit first, then add the pre-allocation change in a second commit.

PR #309, opened 2026-08-09
perf(metal): reduce buffer size binding allocations

KEINOS:
## Summary

This PR follows up on the [minor performance note from the review of #306](https://github.com/gogpu/wgpu/pull/306#pullrequestreview-4891380026). `ComputePassEncoder.bindBufferSizes()` sends runtime-array sizes before each direct or indirect dispatch.

I first tried reusing a pre-allocated Go slice. Measurements showed that the original local slice does not escape. The reusable slice added one allocation per pass. A second version used an inline array, but it did not reduce allocations and made one benchmark case slightly slower.

Profiling showed that most allocations came from preparing the generic Objective-C call. This change prepares the call interface once and uses a specialized `setBytes:length:atIndex:` path. The buffer-size values and binding behavior stay unchanged.

## Testing

The first commit adds functional Metal tests without allocation assertions. They cover repeated direct dispatches with compatible pipeline and bind group changes, plus indirect dispatch with two runtime-sized arrays.

```console
go test -tags integration -run '^(TestArrayLength|TestBufferSizes)' -count=5 -v .
go test -race -count=1 ./hal/metal/...
go test -count=1 ./...
go build ./...
```

The benchmark uses a real Metal compute encoder with the same inputs before and after the implementation:

```console
go test -run='^$' -bench='^BenchmarkComputePassEncoderBindBufferSizes$' -benchmem -count=10 ./hal/metal
```

Apple M4, Go 1.26.5, 10 samples:

| Dispatches per pass | Time | Memory | Allocations |
| --- | ---: | ---: | ---: |
| 1 | 181.1 → 133.7 ns/op (-26.15%) | 320 → 184 B/op (-42.50%) | 9 → 6 allocs/op (-33.33%) |
| 8 | 1.462 → 1.080 µs/op (-26.10%) | 2561 → 1472 B/op (-42.52%) | 72 → 48 allocs/op (-33.33%) |
| 64 | 11.780 → 8.620 µs/op (-26.83%) | 20489 → 11781 B/op (-42.50%) | 576 → 384 allocs/op (-33.33%) |

The `benchstat` comparison shows lower execution time, memory use, and allocation counts in all three cases.

## Scope

- Metal compute passes only.
- No public API changes.
- No render-path changes or unrelated lint cleanup.

### DIFF
--- a/arraylength_repro_test.go
+++ b/arraylength_repro_test.go
@@ -45,6 +45,28 @@ const noArrayLengthShader = `
 fn main() {}
 `
 
+const twoBufferSizesShader = `
+@group(0) @binding(0) var<storage, read_write> first: array<u32>;
+@group(0) @binding(1) var<storage, read_write> second: array<u32>;
+
+@compute @workgroup_size(1)
+fn main() {
+    first[0] = arrayLength(&first);
+    first[1] += 1u;
+    second[0] = arrayLength(&second);
+    second[1] += 1u;
+}
+`
+
+const oneBufferSizeShader = `
+@group(0) @binding(0) var<storage, read_write> first: array<u32>;
+
+@compute @workgroup_size(1)
+fn main() {
+    first[2] = arrayLength(&first);
+}
+`
+
 // arrayLengthElems is the element count bound to the storage binding.
 const arrayLengthElems = 4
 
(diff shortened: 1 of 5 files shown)


kolkov (approved):
LGTM — Metal perf optimization, zero conflicts with ADR-060 work. CI 17/17 green, codecov pass.

PR #311, opened 2026-08-10
fix(core): own surface configuration state

besmpl:
## Summary

- copy surface configuration before passing it to HAL or storing it in core state
- return a defensive copy from `Surface.Config`
- preserve the previous configuration when an initial or subsequent HAL configure attempt fails
- add ownership, failure, and concurrency regressions

## Why

`Surface.Configure` stored the caller's `*hal.SurfaceConfiguration` directly, and `Config()` returned the same internal pointer. Mutating either pointer silently changed core dimensions/format without reconfiguring HAL and raced with concurrent core reads.

`hal.SurfaceConfiguration` contains only scalar fields, so value copying completely establishes ownership without a deep-copy dependency.

## Verification

- baseline input mutation changes the stored config
- baseline mutation of `Config()` changes subsequent reads
- baseline concurrent mutation reproduces a race
- failed configure/reconfigure calls leave committed state unchanged
- unconfigured `Config()` remains nil
- focused ownership tests repeated 100 times
- `go test ./core`
- `go test -race ./core`
- `go test ./...`
- `go build ./...`
- Linux, Windows, Darwin, and WASM builds

The full repository race suite still reports the pre-existing `hal/noop.Queue.Submit` counter race; the changed core surface tests are race-clean.

### DIFF
--- a/core/surface.go
+++ b/core/surface.go
@@ -54,7 +54,8 @@ func (s *Surface) SetPrepareFrame(fn PrepareFrameFunc) {
 // configured, it will be reconfigured with the new settings.
 //
 // After Configure, the surface enters the Configured state and is ready
-// to acquire textures.
+// to acquire textures. The configuration is copied; the surface does not
+// retain the caller's pointer.
 func (s *Surface) Configure(device *Device, config *hal.SurfaceConfiguration) error {
 	s.mu.Lock()
 	defer s.mu.Unlock()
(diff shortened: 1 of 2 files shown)


kolkov (approved):
Verified all 4 claims against source code:

1. **Pointer aliasing** — confirmed: old code stored caller's `*hal.SurfaceConfiguration` directly. Fix: value copy before store. Correct.
2. **Config() mutation** — confirmed: old code returned internal pointer. Fix: returns copy. Correct.
3. **Failed configure** — confirmed: assignment happens only after HAL success. Previous state preserved on error.
4. **Value copy sufficient** — confirmed: `hal.SurfaceConfiguration` is all scalars (Width, Height, Format, Usage, PresentMode, AlphaMode, EnableDamagePresent). No deep copy needed.

Thread safety correct — both Configure and Config operate under existing `s.mu`.

No conflict with ADR-060 changes (core layer, not surface_native.go). The `swapchainTexture` cache reads from `Config()` — returning a copy is strictly better for this path.

Tests thorough: ownership, return-copy, race, and failure paths all covered.

PR #312, opened 2026-08-10
fix: track native texture copy usage

besmpl:
## Summary

- record CopySrc/CopyDst texture usage for all three native texture-copy APIs
- preflight multi-resource usage changes atomically before mutating command scopes
- retain buffer endpoints until submitted GPU work retires
- preserve explicit-transition replacement semantics
- add scope, conflict, released-resource, lifetime, barrier, and guard-path regressions

## Why

The native encoder tracked buffer usage for texture↔buffer copies but never recorded texture usage. Texture↔texture copies recorded neither endpoint. Submit-time tracking therefore could not generate transitions from an already tracked Resource or ColorTarget state to the transfer layouts required by Vulkan/DX12.

Copy commands now validate every endpoint first and commit their scopes only when the complete operation is valid. Failed operations leave scopes, resource refs, submit-validation sets, and the HAL command stream untouched. Source/destination buffers are retained through queue retirement.

## Verification

- archived baseline fails all three copy-scope cases, released-resource cases, lifetime cases, and seven atomic-conflict cases
- tracked Resource/ColorTarget → CopySrc/CopyDst transitions are generated
- explicit transitions remain compatible with following copies
- conflicting same-texture copy roles are rejected without partial state
- buffer refcounts remain live through submission and drop at completion
- copy-usage guard, compatibility, conflict, and explicit-transition branches are exercised
- Codecov patch coverage: 100%; all modified and coverable lines covered
- `go test ./...`
- `go test -race . ./core ./core/track`
- `go build ./...`
- `go vet . ./core ./core/track`
- Rust-tag, WASM, Linux, and Windows builds
- golangci-lint v2.7.2 changed lines: 0 issues

## Known prerequisite dependencies

This PR intentionally fixes command-scope recording, not the complete queue transition engine. It remains draft because correct emitted transitions require coordinated state-model and submission-order work:

1. **First-use and lifecycle state.** Newly created textures are not registered as `Uninitialized`. Registration cannot stand alone: tracker state must be removed exactly once before tracker-index reuse, swapchain acquire/present must account for each physical image, and pending `WriteTexture`/explicit-transition paths must keep the tracker synchronized.
2. **Command-buffer ordering and atomicity.** Submission currently gathers transitions into one preamble before all command buffers. Correct ordering needs scopes interleaved with pending-write buffers and each HAL command-buffer segment (one public command buffer may own several HAL buffers), plus transactional tracker commit/rollback so an encoder or submit failure cannot advance CPU state without the corresponding GPU transition.

Until those dependencies are solved, this PR's claims are limited to correct atomic copy-scope and resource-lifetime recording.

## Review follow-up (2026-08-12)

- `CopyBufferToBuffer` now validates both endpoint usages atomically before mutating scope state, retaining refs, or recording HAL commands.
- Added success and conflict regressions for both buffer endpoints.
- Documented the unconditional/preflight-only contract of `ReplaceUsage` and added the matching wgpu-core `command/transfer.rs` references.

The PR intentionally remains draft for the broader queue-transition prerequisites already listed above.

### DIFF
--- a/copy_usage_coverage_test.go
+++ b/copy_usage_coverage_test.go
@@ -14,7 +14,7 @@ func TestCopyUsageHelpersSkipUntrackedResources(t *testing.T) {
 	t.Parallel()
 
 	encoder := &CommandEncoder{}
-	if !encoder.recordBufferUsage(nil, track.BufferUsesCopySrc) {
+	if !encoder.recordCopyBufferUsages([]copyBufferUsage{{usage: track.BufferUsesCopySrc}}) {
 		t.Fatal("nil buffer should be ignored")
 	}
 	if !encoder.recordCopyUsages(nil, nil, track.BufferUsesNone) {
@@ -59,7 +59,7 @@ func TestCopyUsagePreflightGuardBranches(t *testing.T) {
 		t.Fatalf("CreateCommandEncoder: %v", err)
 	}
 	defer encoder.DiscardEncoding()
-	if !encoder.recordBufferUsage(nil, track.BufferUsesCopySrc) {
+	if !encoder.recordCopyBufferUsages([]copyBufferUsage{{usage: track.BufferUsesCopySrc}}) {
 		t.Fatal("nil buffer with live core encoder should be ignored")
 	}
 
@@ -133,8 +133,8 @@ func TestCopyUsagePreflightGuardBranches(t *testing.T) {
 		t.Fatalf("compatible buffer preflight = (usage %v, tracked %v, err %v), want Vertex|Uniform", usage, tracked, err)
 	}
 	encoder.core.Mutable().BufferScope().ReplaceUsage(bufferIndex, track.BufferUsesStorageWrite)
-	if encoder.recordBufferUsage(buffer.core, track.BufferUsesCopySrc) {
-		t.Fatal("recordBufferUsage accepted incompatible usage")
+	if encoder.recordCopyBufferUsages([]copyBufferUsage{{buffer: buffer.core, usage: track.BufferUsesCopySrc}}) {
+		t.Fatal("recordCopyBufferUsages accepted incompatible usage")
 	}
 }
 
--- a/copy_usage_coverage_test.go
+++ b/copy_usage_coverage_test.go
@@ -1,0 +1,175 @@
+//go:build !rust && !(js && wasm)
+
+package wgpu
+
+import (
+	"testing"
+
+	"github.com/gogpu/gputypes"
+	"github.com/gogpu/wgpu/core"
+	"github.com/gogpu/wgpu/core/track"
+)
+
+func TestCopyUsageHelpersSkipUntrackedResources(t *testing.T) {
+	t.Parallel()
+
+	encoder := &CommandEncoder{}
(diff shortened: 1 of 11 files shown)


kolkov (commented):
Excellent work — this fills a gap our own validation agents identified (copy commands not recording texture usage). The atomic preflight pattern (prepare → validate all → commit) is clean and well-tested.

**Code quality: enterprise level.** 1,102 LOC tests, 100% patch coverage, 7 atomic-conflict scenarios, full lifecycle test through Submit. Consistent with our ADR-060 patterns.

**ADR-060 compatibility:** changes to `core/track/texture.go` and `core/track/buffer.go` are purely additive (`ReplaceUsage`). `encoder_native.go` restructures copy commands — consistent with our tracker direction.

**Prerequisites (from your description):** confirmed real but pre-existing, not introduced by this PR. The scope recording is independently correct.

Three items to address:

1. **`CopyBufferToBuffer` consistency.** The PR changes `recordBufferUsage` return type to `bool` but `CopyBufferToBuffer` (lines 181-182) still ignores the return value and proceeds to the HAL call on conflict. The other three copy commands use the new atomic pattern. Either update `CopyBufferToBuffer` to match, or note the exclusion in the PR description.

2. **Rust reference comments.** `RecordTextureUsage` and `ReplaceTextureUsage` in `core/command.go` should reference `wgpu-core command/transfer.rs` for consistency with `RecordBufferUsage`.

3. **`ReplaceUsage` doc comment.** A note that it is intended for use after preflight validation only (unconditional write, no conflict check) would clarify the contract.

None of these are architectural — happy to approve once item 1 is addressed.

kolkov (approved):
All three items from the previous review are addressed:

1. **CopyBufferToBuffer atomic preflight** — now uses `recordCopyBufferUsages` with prepare → validate → commit pattern, consistent with the other 3 copy commands. Failed copies no longer clone ResourceRefs.

2. **Rust reference comments** — added on `RecordTextureUsage` and `ReplaceTextureUsage`.

3. **ReplaceUsage doc comment** — added on both `BufferUsageScope.ReplaceUsage` and `TextureUsageScope.ReplaceUsage`, clarifying post-preflight intent.

Test quality is excellent — 5 test files covering atomicity invariants (7 sub-tests for all copy command combinations), resource lifecycle through encoding/submit/retirement, guard branches, and edge cases.

LGTM.

kolkov:
Hey @besmpl — all review items addressed, CI green. Please double-check everything on your side and mark "Ready for review" when you're confident. We'll merge with `--squash`.

kolkov:
Merged — thank you for the excellent work on this. The atomic preflight pattern across all 4 copy commands is exactly what we needed, and the test coverage (atomicity invariants, refcount lifecycle, guard branches) is enterprise quality.

Apologies for not waiting for you to remove draft — this has been important for our tracker pipeline and we wanted to get it into the next release. All review items were addressed and CI was green, so we went ahead with the squash merge.

Also appreciated the extra fixes bundled in: `CopyBufferToBuffer` now properly reports an error on released buffers (was silently returning), and `CopyBufferToTexture` caches `halSrc` with a nil check (was calling `halBuffer()` unchecked at the call site).

PR #318, opened 2026-08-18
fix: add writable mapped-range API across backends

tarmo888:
## Summary

Add a common writable mapped-range API to the native `wgpu` backend:

```go
BytesMut() []byte
Flush() error
```

The browser and Rust backends already expose equivalent behavior. This change brings the native backend into API parity so higher-level libraries such as `g3d` can write mapped buffer data without build-tag-specific code.

## Motivation

`MappedRange.Bytes()` is currently suitable for reading, but higher-level code that writes mapped buffer contents needs an explicit writable-range API.

The browser backend uses a Go-side staging slice for mapped writes and requires `Flush()` to copy modified data back to the JavaScript `ArrayBuffer`. The Rust backend exposes a direct writable mapping. Native mappings also expose direct writable memory.

Without a common API, consumers need platform-specific implementations for the same geometry or buffer upload code.

## Changes

- Added `MappedRange.BytesMut()` to the native backend.
- Added `MappedRange.Flush()` to the native backend.
- Implemented native `BytesMut()` as an alias of `Bytes()`.
- Implemented native `Flush()` as a no-op because native mappings are directly device-visible.
- Added a regression test

## Related

This change is a prerequisite for the corresponding `gogpu/g3d` update that switches geometry uploads to the common `BytesMut()` and `Flush()` API.

### DIFF
--- a/mapped_range.go
+++ b/mapped_range.go
@@ -77,6 +77,19 @@ func (m *MappedRange) Bytes() []byte {
 	return unsafe.Slice((*byte)(m.data), m.size) //nolint:gosec // range validated at MappedRange creation
 }
 
+// BytesMut returns a mutable byte slice for writing into the mapped region.
+// On native backend, this returns the same underlying data as Bytes() since
+// the mapped range is already writable through the direct mapping.
+func (m *MappedRange) BytesMut() []byte {
+	return m.Bytes()
+}
+
+// Flush writes the cached data back. On native backend, data is written through
+// the direct pointer, so this is a no-op.
+func (m *MappedRange) Flush() error {
+	return nil
+}
+
 // Len returns the size of the mapped range in bytes.
 func (m *MappedRange) Len() int {
 	if m == nil {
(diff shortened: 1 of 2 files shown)


kolkov (approved):
Clean PR — fills the native backend gap for `BytesMut()`/`Flush()` that browser and Rust backends already have. CI 16/16 pass.

Implementation is correct: native `BytesMut()` = alias of `Bytes()`, `Flush()` = no-op. This matches the Rust backend exactly (`mapped_range_rust.go` uses the same pattern).

Test follows existing conventions (`newDevice` + `requireHAL` + `MappedAtCreation`).

### Design note (non-blocking, for awareness)

`Flush()` as no-op is correct for DX12, Metal, GLES, Software, and Noop — all return `IsCoherent: true` from `MapBuffer`. Vulkan CAN return `IsCoherent: false` for non-coherent memory types, where CPU writes aren't GPU-visible without `vkFlushMappedMemoryRanges`. However, this is a pre-existing gap — `MappedRange` doesn't store the coherency flag, and `Bytes()` already allows writes with no flush mechanism. The Vulkan `UnmapBuffer` handles the flush. A proper `Flush()` for non-coherent Vulkan would need plumbing `IsCoherent` into `MappedRange` — that's a separate enhancement.

LGTM.

PR #323, opened 2026-08-27
Fix Vulkan array texture uploads

dvoyni:
`Queue.WriteTexture` currently targets array layer 0 on native Vulkan even when `ImageCopyTexture.Origin.Z` selects a nonzero layer.

Two conversions contributed to the issue:

- pending-write barriers hardcoded `BaseArrayLayer: 0` and `ArrayLayerCount: 1`
- Vulkan buffer/image copies hardcoded `BaseArrayLayer: 0` and mapped `Origin.Z` to `ImageOffset.Z`, which is correct for 3D textures but not 2D texture arrays

This change makes pending-write barriers target the requested array subresources and makes Vulkan copy conversion dimension-aware. Non-3D textures map `Origin.Z`/`DepthOrArrayLayers` to array layers with a zero depth offset; 3D textures retain the existing depth behavior.

Regression tests cover nonzero 2D array layers and 3D depth for both pending barriers and Vulkan copy conversion.

Validation: `go test ./...`

### DIFF
--- a/hal/vulkan/command.go
+++ b/hal/vulkan/command.go
@@ -614,7 +614,7 @@ func (e *CommandEncoder) CopyBufferToBuffer(src, dst hal.Buffer, regions []hal.B
 // convertBufferImageCopyRegions converts HAL BufferTextureCopy regions to Vulkan BufferImageCopy.
 // The format parameter is the texture format, used to determine block copy size
 // for correct bytes-to-texels conversion of bufferRowLength.
-func convertBufferImageCopyRegions(regions []hal.BufferTextureCopy, format gputypes.TextureFormat) []vk.BufferImageCopy {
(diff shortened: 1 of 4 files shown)


kolkov (approved):
Clean bug fix with correct Rust wgpu reference parity. CI 15/15 pass.

**Rust comparison:** Rust wgpu splits `Origin.Z` into `array_layer` vs `origin.z` upstream in `wgpu-core/src/command/transfer.rs:249-266` (`extract_texture_selector`). Our Go architecture carries `Origin3D.Z` directly to the HAL, so splitting at the HAL conversion point (as this PR does) is the correct Go-idiomatic equivalent. The `!= TextureDimension3D` check matches Rust's three-way match exactly.

**Additional correctness improvement:** The `Aspect` field change from hardcoded `TextureAspectAll` to `dst.Aspect` matches Rust behavior and is required for correct depth-stencil subresource transitions.

**DX12 consistency:** The DX12 backend already uses the same `!= TextureDimension3D` pattern (`hal/dx12/copy_plan.go:302-307`), confirming the approach.

LGTM.

Issue #330, opened 2026-08-28
feat: expose CreateQuerySet + CreateRenderBundleEncoder in public API


kolkov:
## Summary

`Device.CreateQuerySet` and `Device.CreateRenderBundleEncoder` are fully implemented on all 6 HAL backends but have no public API wrapper. Users must access HAL directly to use these features.

## What's ready (all 6 backends)

**CreateQuerySet:**
- Vulkan: `hal/vulkan/query.go`
- DX12: `hal/dx12/query.go`
- Metal: `hal/metal/device.go`
- GLES: `hal/gles/device.go`
- Software: `hal/software/device.go` (returns ErrTimestampsNotSupported)
- Noop: `hal/noop/device.go` (returns ErrTimestampsNotSupported)

**CreateRenderBundleEncoder:** implemented on all 6 backends.

## What's missing

Two methods in root package (`device_native.go`, ~30 LOC each):

```go
func (d *Device) CreateQuerySet(desc *QuerySetDescriptor) (*QuerySet, error)
func (d *Device) CreateRenderBundleEncoder(desc *RenderBundleEncoderDescriptor) (*RenderBundleEncoder, error)
```

Plus corresponding public types if not yet in the root package.

## Impact

Without `CreateQuerySet`, users cannot do timestamp or occlusion queries through the public API. This is a gap for profiling and performance measurement.

Issue #333, opened 2026-08-28
feat: validation Phase C — expand to ~70% Rust wgpu-core parity


kolkov:
## Summary

Expand WebGPU validation coverage from ~45% to ~70% of Rust wgpu-core checks. Currently 55 validation functions across `core/validate.go` (26), `internal/raytracing/validate.go` (17), and `core/indirect_validation.go` (12). Key gaps: feature gates, texture usage conflicts, render pass compatibility.

### Feature gate checks (highest impact)
None of the 25 feature flags are validated at usage time:
- `FeatureTimestampQuery` — not checked on `CreateQuerySet`
- `FeatureMultiDrawIndirect` — not checked on multi-draw calls
- `FeaturePushConstants` — not checked on PipelineLayout with push constants
- `FeatureShaderF16` — not checked on shader module with f16
- `FeatureRayQuery` — checked in `internal/raytracing/` (9 checks), but NOT in core/

### Texture usage conflicts
- Complete conflict matrix (CopySrc+CopyDst mutual exclusion in same pass)
- Storage texture read-write conflict detection
- Multisampled texture usage restrictions

### Render pass compatibility
- Color attachment format mismatch with pipeline
- Depth/stencil format mismatch with pipeline
- Sample count consistency between all attachments

### Buffer usage
- `MAP_READ` + `MAP_WRITE` mutual exclusion validation
- Storage buffer alignment checks
- Uniform buffer size limits per binding

## References

- Rust wgpu-core: `command/render.rs`, `command/compute.rs`, `binding_model.rs`, `device/resource.rs`
- Our validation: `core/validate.go`, `internal/raytracing/validate.go`, `core/indirect_validation.go`
- Validation Phases A+B already done (crash prevention + correctness, ~45% coverage)

lkmavi:
@kolkov Let's create a task in our internal special to add additional tests

kolkov:
Good idea. The validation tests should cover all feature gate checks (25 features, none currently validated at usage time) and texture usage conflict matrix. Feel free to start with feature gates — they're the highest impact gap.

kolkov:
Resolved in PR #347 (v0.34.2+). Validation Phase C: 15+ feature gates, texture usage validation, indirect count with Vulkan native vkCmdDrawIndirectCount, ~80 tests. @lkmavi

PR #337, opened 2026-08-30
feat: expose query sets and render bundle encoders

gusevgrishaem1:
## Summary

This PR adds public query set types and render bundle encoders.

Related to #330


kolkov (changes requested):
Thanks for the PR! Good structural foundation — build tags, lifecycle pattern (atomic.Bool + idempotent Release), and HAL delegation are all correct.

A few issues to address before merge:

## 1. QueryType duplication

`QueryType` is already defined in `hal/descriptor.go:547-555` with identical values. The PR adds a second independent definition in `types.go`, then casts between them in `toHAL()` via `hal.QueryType(d.Type)`.

In Rust wgpu, `QueryType` lives in wgpu-types (the shared types crate) — single source of truth ([wgpu-types/src/lib.rs:450](https://github.com/gfx-rs/wgpu/blob/trunk/wgpu-types/src/lib.rs#L450)). Our equivalent is `gputypes`, but for now a type alias works:

```go
type QueryType = hal.QueryType

const (
    QueryTypeOcclusion = hal.QueryTypeOcclusion
    QueryTypeTimestamp  = hal.QueryTypeTimestamp
)
```

This eliminates the duplication and the fragile cast.

## 2. Incomplete API — QuerySet is unusable without companion methods

Creating a `QuerySet` is only half the story. The W3C WebGPU spec ([index.bs:11239-11295](https://www.w3.org/TR/webgpu/#dom-gpucommandencoder-resolvequeryset)) requires companion methods to actually *use* queries:

| Method | Spec | Our HAL | Public API |
|--------|------|---------|------------|
| `CommandEncoder.resolveQuerySet()` | Required | ✅ `hal/command.go:62-68` | ❌ Missing |
| `RenderPassDescriptor.timestampWrites` | Required | ✅ `hal/descriptor.go:481-482` | ❌ Missing |
| `RenderPassDescriptor.occlusionQuerySet` | Required | ❌ Not in HAL | ❌ Missing |
| `RenderPass.beginOcclusionQuery()` | Required | ❌ Not in HAL | ❌ Missing |
| `RenderPass.endOcclusionQuery()` | Required | ❌ Not in HAL | ❌ Missing |

Without at least `resolveQuerySet`, a user creates a QuerySet but can never read results from it.

Suggestion: either add `CommandEncoder.ResolveQuerySet()` in this PR (the HAL implementation is ready on all backends), or add a doc comment stating "Phase 1 — companion methods in follow-up PR" with a link to #330.

## 3. Incomplete API — RenderBundle is unusable without ExecuteBundles

Same pattern: `RenderBundleEncoder.Finish()` produces a `RenderBundle`, but there's no way to execute it. The spec requires `RenderPass.executeBundles()` ([index.bs:13935-13948](https://www.w3.org/TR/webgpu/#dom-gpurenderpassencoder-executebundles)), and our HAL already implements it (`hal/command.go:149-151`, all 6 backends).

Suggestion: add `RenderPassEncoder.ExecuteBundles(bundles ...*RenderBundle)` to the public API in this PR.

## 4. Finish() should accept a descriptor

Both the W3C spec and Rust wgpu pass a descriptor to finish:

- Spec: `GPURenderBundle finish(optional GPURenderBundleDescriptor descriptor = {})` ([index.bs:14075](https://www.w3.org/TR/webgpu/#dom-gpurenderbundleencoder-finish))
- Rust: `pub fn finish(self, desc: &RenderBundleDescriptor) -> RenderBundle` ([render_bundle_encoder.rs:76](https://github.com/gfx-rs/wgpu/blob/trunk/wgpu/src/api/render_bundle_encoder.rs#L76))

The descriptor contains at minimum a label for the resulting bundle. Suggestion:

```go
func (e *RenderBundleEncoder) Finish(desc *RenderBundleDescriptor) (*RenderBundle, error)
```

Returning `(*RenderBundle, error)` is more Go-idiomatic than returning nil on double-finish.

## 5. Encoder methods need guard after Finish()

The W3C spec defines an encoder state machine ([index.bs:10455-10491](https://www.w3.org/TR/webgpu/#encoder-state)): commands on an "ended" encoder must generate a validation error. Currently only `Finish()` checks `e.finished` — the 6 draw/state methods (`SetPipeline`, `SetBindGroup`, `SetVertexBuffer`, `SetIndexBuffer`, `Draw`, `DrawIndexed`) pass straight through to HAL after finish.

Go doesn't have Rust's ownership semantics that prevent this at compile time, so a runtime check is needed:

```go
func (e *RenderBundleEncoder) SetPipeline(pipeline *RenderPipeline) {
    if e.finished.Load() {
        return // or log warning
    }
    e.hal.SetPipeline(pipeline.hal)
}
```

---

Overall the code quality is good — you clearly understand the wgpu architecture. These changes will make the API complete and usable. Happy to help if you have questions!

gusevgrishaem1:
Hi! Thanks for the review! This is my first open-source contribution, so I really appreciate your feedback. I’ll update the MR according to your comments soon.

> Thanks for the PR! Good structural foundation — build tags, lifecycle pattern (atomic.Bool + idempotent Release), and HAL delegation are all correct.
> 
> A few issues to address before merge:
> 
> ## 1. QueryType duplication
> `QueryType` is already defined in `hal/descriptor.go:547-555` with identical values. The PR adds a second independent definition in `types.go`, then casts between them in `toHAL()` via `hal.QueryType(d.Type)`.
> 
> In Rust wgpu, `QueryType` lives in wgpu-types (the shared types crate) — single source of truth ([wgpu-types/src/lib.rs:450](https://github.com/gfx-rs/wgpu/blob/trunk/wgpu-types/src/lib.rs#L450)). Our equivalent is `gputypes`, but for now a type alias works:
> 
> ```go
> type QueryType = hal.QueryType
> 
> const (
>     QueryTypeOcclusion = hal.QueryTypeOcclusion
>     QueryTypeTimestamp  = hal.QueryTypeTimestamp
> )
> ```
> 
> This eliminates the duplication and the fragile cast.
> 
> ## 2. Incomplete API — QuerySet is unusable without companion methods
> Creating a `QuerySet` is only half the story. The W3C WebGPU spec ([index.bs:11239-11295](https://www.w3.org/TR/webgpu/#dom-gpucommandencoder-resolvequeryset)) requires companion methods to actually _use_ queries:
> 
> Method	Spec	Our HAL	Public API
> `CommandEncoder.resolveQuerySet()`	Required	✅ `hal/command.go:62-68`	❌ Missing
> `RenderPassDescriptor.timestampWrites`	Required	✅ `hal/descriptor.go:481-482`	❌ Missing
> `RenderPassDescriptor.occlusionQuerySet`	Required	❌ Not in HAL	❌ Missing
> `RenderPass.beginOcclusionQuery()`	Required	❌ Not in HAL	❌ Missing
> `RenderPass.endOcclusionQuery()`	Required	❌ Not in HAL	❌ Missing
> Without at least `resolveQuerySet`, a user creates a QuerySet but can never read results from it.
> 
> Suggestion: either add `CommandEncoder.ResolveQuerySet()` in this PR (the HAL implementation is ready on all backends), or add a doc comment stating "Phase 1 — companion methods in follow-up PR" with a link to #330.
> 
> ## 3. Incomplete API — RenderBundle is unusable without ExecuteBundles
> Same pattern: `RenderBundleEncoder.Finish()` produces a `RenderBundle`, but there's no way to execute it. The spec requires `RenderPass.executeBundles()` ([index.bs:13935-13948](https://www.w3.org/TR/webgpu/#dom-gpurenderpassencoder-executebundles)), and our HAL already implements it (`hal/command.go:149-151`, all 6 backends).
> 
> Suggestion: add `RenderPassEncoder.ExecuteBundles(bundles ...*RenderBundle)` to the public API in this PR.
> 
> ## 4. Finish() should accept a descriptor
> Both the W3C spec and Rust wgpu pass a descriptor to finish:
> 
> * Spec: `GPURenderBundle finish(optional GPURenderBundleDescriptor descriptor = {})` ([index.bs:14075](https://www.w3.org/TR/webgpu/#dom-gpurenderbundleencoder-finish))
> * Rust: `pub fn finish(self, desc: &RenderBundleDescriptor) -> RenderBundle` ([render_bundle_encoder.rs:76](https://github.com/gfx-rs/wgpu/blob/trunk/wgpu/src/api/render_bundle_encoder.rs#L76))
> 
> The descriptor contains at minimum a label for the resulting bundle. Suggestion:
> 
> ```go
> func (e *RenderBundleEncoder) Finish(desc *RenderBundleDescriptor) (*RenderBundle, error)
> ```
> 
> Returning `(*RenderBundle, error)` is more Go-idiomatic than returning nil on double-finish.
> 
> ## 5. Encoder methods need guard after Finish()
> The W3C spec defines an encoder state machine ([index.bs:10455-10491](https://www.w3.org/TR/webgpu/#encoder-state)): commands on an "ended" encoder must generate a validation error. Currently only `Finish()` checks `e.finished` — the 6 draw/state methods (`SetPipeline`, `SetBindGroup`, `SetVertexBuffer`, `SetIndexBuffer`, `Draw`, `DrawIndexed`) pass straight through to HAL after finish.
> 
> Go doesn't have Rust's ownership semantics that prevent this at compile time, so a runtime check is needed:
> 
> ```go
> func (e *RenderBundleEncoder) SetPipeline(pipeline *RenderPipeline) {
>     if e.finished.Load() {
>         return // or log warning
>     }
>     e.hal.SetPipeline(pipeline.hal)
> }
> ```
> 
> Overall the code quality is good — you clearly understand the wgpu architecture. These changes will make the API complete and usable. Happy to help if you have questions!

Hi! Thanks for the review! This is my first open-source contribution, so I really appreciate the feedback. I’ll address the comments and update the MR soon.

gusevgrishaem1:
Hi! I've made the requested changes. Could you please take another look when you have time? Thanks!

lkmavi:
Hi @gusevgrishaem1 — thanks again for #337 and for addressing @kolkov's first-round feedback. The render-bundle side and companion APIs (`ResolveQuerySet`, `ExecuteBundles`, timestamp writes, Finish guards) look solid, and the test file is strong.

We opened **#347** ([Validation Phase C — #333](https://github.com/gogpu/wgpu/pull/347)) in parallel, and it **overlaps with the QuerySet half of this PR**. To avoid merge conflicts and duplicated API surface, we'd like to coordinate before you invest more time in the query-set path.

### What #347 adds on the QuerySet side

| Area | #337 (this PR) | #347 |
|------|----------------|------|
| `Device.CreateQuerySet` | ✅ thin HAL wrapper | ✅ + `core.ValidateQuerySetDescriptor` (VAL-C24) |
| Feature gates | ❌ | ✅ `FeatureTimestampQuery` / `FeaturePipelineStatisticsQuery` via `RequireFeature` |
| `QuerySet` type | `{hal, device, released}` | `core.QuerySet` integration |
| `Type()` / `Count()` | ❌ | ✅ |
| `Release()` lifecycle | immediate `DestroyQuerySet` | **DestroyQueue-deferred** (ADR-056, same as Buffer/BindGroup) |
| Tests | capture HAL mocks in `query_renderbundle_native_test.go` | `core/query_validate_test.go`, `query_native_test.go`, registry-driven `featureGateCases`, `wgpu_feature_gate_test.go` smoke |
| Docs / CI | — | `docs/VALIDATION-TESTING.md`, full CI + codecov patch |

#347 is **CI-green** and targets main as part of validation work (#333), which also covers #330's validation requirements for query sets.

### What #337 still has that #347 does not (yet)

These are valuable and **not duplicated** in #347:

- `CommandEncoder.ResolveQuerySet`
- `RenderPassDescriptor.TimestampWrites`
- Full **RenderBundle** public API (`CreateRenderBundleEncoder`, `Finish(desc)`, `ExecuteBundles`, post-Finish guards)

We definitely want those — ideally as a follow-up once QuerySet foundation is settled.

### Ask

**Please hold off on further QuerySet changes in this branch** until #347 lands (or we agree on merge order). After that:

1. Rebase #337 onto `main`
2. **Drop** the QuerySet / `CreateQuerySet` / `query_native.go` parts (already in #347)
3. **Keep** render-bundle + `ResolveQuerySet` + timestamp writes — we can review that slice cleanly without conflict

Happy to pair on the rebase split or open a short coordination issue if that helps. Thanks for your patience 🙏

lkmavi:
Hi @gusevgrishaem1 — thanks again for the patience 🙏

**#347 is now merged into `main`**, so we can unblock #337. GitHub shows this PR as **conflicting** with `main`. Next step is a rebase/merge plus a small adaptation so we keep your APIs without duplicating QuerySet.

Really appreciate the work you already put in — especially completing the companion methods and RenderBundle surface after the first review round. Once this is synced with `main`, the remaining slice should be clean and close to merge-ready.

---

### Goal after update

| Keep from #337 | Already on `main` (#347) — do **not** reintroduce |
|---|---|
| `renderbundle_native.go` | `query_native.go` (use **main**’s version) |
| `Device.CreateRenderBundleEncoder` | `Device.CreateQuerySet` / `QuerySet` / `QuerySetDescriptor` |
| `RenderPassEncoder.ExecuteBundles` | feature gates + `core.ValidateQuerySetDescriptor` |
| `CommandEncoder.ResolveQuerySet` | QuerySet-only unit tests already covered on main |
| `RenderPassDescriptor.TimestampWrites` | |
| RenderBundle / Resolve / Timestamp tests | |

---

### Step-by-step

#### 1) Sync your fork branch with upstream `main`

```bash
git remote add upstream https://github.com/gogpu/wgpu.git   # skip if already added
git fetch upstream
git checkout feat/public-query-set-render-bundle
git merge upstream/main
# alternative: git rebase upstream/main
```

Expect conflicts mainly in:
- `query_native.go` (**added in both**)
- `descriptor.go`
- `device_native.go`
- `renderpass_native.go`

#### 2) `query_native.go` — take **main**’s implementation

Keep the #347 version:
- `core *core.QuerySet` (not a raw `hal.QuerySet` field)
- `Type()` / `Count()`
- DestroyQueue-deferred `Release()`
- `CreateQuerySet` lives in this file on main

Quick check: the file should import `github.com/gogpu/wgpu/core` and call `core.NewQuerySet(...)`.

#### 3) `device_native.go` — drop duplicate `CreateQuerySet`, keep RenderBundle

- **Remove** your `CreateQuerySet` (already on main in `query_native.go`)
- **Keep** `CreateRenderBundleEncoder`

#### 4) `descriptor.go` — don’t duplicate QuerySet types

- Keep **main**’s `QuerySetDescriptor` / `QueryType` aliases
- From your PR, **add**:
  - `RenderBundleEncoderDescriptor`
  - `RenderBundleDescriptor`
  - `RenderPassTimestampWrites`
  - `TimestampWrites` on `RenderPassDescriptor` + `toHAL()` wiring

#### 5) Adapt `ResolveQuerySet` + `TimestampWrites` to main’s `QuerySet`

On main, `QuerySet` no longer has `.hal` or `released atomic.Bool`. Add a helper similar to `Buffer.halBuffer()`:

```go
func (qs *QuerySet) halQuerySet() hal.QuerySet {
	if qs == nil || qs.core == nil || qs.device == nil {
		return nil
	}
	guard := qs.device.core.SnatchLock().Read()
	defer guard.Release()
	return qs.core.Raw(guard)
}
```

Then update call sites:

```go
// before (your PR)
if querySet.released.Load() || querySet.hal == nil { ... }
raw.ResolveQuerySet(querySet.hal, firstQuery, queryCount, dst, dstOffset)

// after (main QuerySet)
if querySet.released || querySet.halQuerySet() == nil { ... }
raw.ResolveQuerySet(querySet.halQuerySet(), firstQuery, queryCount, dst, dstOffset)
```

Same idea in `RenderPassDescriptor.toHAL()` for `TimestampWrites.QuerySet`.

#### 6) Fix tests in `query_renderbundle_native_test.go`

- Remove / rewrite tests that construct the old `QuerySet{hal: ...}` shape or re-test `CreateQuerySet` in isolation (main already has `query_native_test.go`)
- Keep tests for RenderBundle Finish / post-Finish guards, `ExecuteBundles`, `ResolveQuerySet`, and `TimestampWrites` → HAL mapping
- For QuerySet objects in those tests, use main’s struct (`core` + `device`) or go through `halQuerySet()`

#### 7) Verify locally, then push

```bash
go test ./... -count=1
go test . -run 'Query|RenderBundle|ResolveQuery|Timestamp' -count=1

git add -A
git commit -m "fix: rebase on main after #347; keep render bundle + resolve/timestamp"
git push origin feat/public-query-set-render-bundle
# if you rebased instead of merging:
# git push --force-with-lease
```

When CI is green, comment here and we’ll do a full re-review of the remaining slice (RenderBundle + ResolveQuerySet + TimestampWrites).

---

If you hit a tricky conflict hunk, a merge/rebase question, or a failing test — ping us right in this PR and we’ll sort it out together. Thanks again for pushing #337 forward; this next sync should make it a really solid, mergeable contribution 🚀

gusevgrishaem1:
Thanks for the feedback! I’ve updated MR according to the "Step-by-step". Could you please take another look?

lkmavi:
Hi @gusevgrishaem1 — thanks for following the step-by-step after #347 🙏

Sync looks good:
- `query_native.go` kept main’s `#347` QuerySet (`core` + deferred `Release`) and only added `halQuerySet()`
- no duplicate `CreateQuerySet` / QuerySet types
- RenderBundle surface, `ResolveQuerySet`, Finish(desc), post-Finish guards, and tests are in good shape
- CI is green

One **blocking** issue before merge:

### TimestampWrites never reach the HAL via `BeginRenderPass`

`RenderPassDescriptor.TimestampWrites` is wired only in `(*RenderPassDescriptor).toHAL()`, but `BeginRenderPass` does **not** call that path. It goes through `convertRenderPassDesc` → `core.RenderPassDescriptor` → `core.convertRenderPassDescriptor`, and neither core type nor converter carries timestamp writes today.

So setting `TimestampWrites` on a public render-pass descriptor has no effect at runtime. `TestRenderPassDescriptorTimestampWrites` only exercises `.toHAL()`, which is why this slipped past CI.

Please wire it through the real path, e.g.:
1. add `TimestampWrites` to `core.RenderPassDescriptor` (mirror the compute-pass pattern)
2. map it in `convertRenderPassDesc`
3. forward it in `core.convertRenderPassDescriptor` to `hal.RenderPassTimestampWrites`
4. add a test that asserts the HAL render-pass begin sees the writes (not only public `toHAL()`)

### Why this wasn’t caught in the earlier review rounds

This isn’t a regression from the #347 rebase — the gap was already present after the first companion-API fix. The first review checklist was effectively “spec / HAL / public API”: HAL already had `TimestampWrites`, so adding the public field + a `toHAL()` mapping looked complete.

But render passes don’t go `desc.toHAL()` → HAL (unlike many `Device.Create*` APIs). The live path is public → **core** → HAL. Public `toHAL()` is effectively unused by `BeginRenderPass`, so the mapping sat on a dead converter. After #347 we focused on QuerySet dedup / `halQuerySet()` / keeping the RenderBundle slice — we didn’t re-walk that call graph. `ResolveQuerySet` / `ExecuteBundles` are fine (they call HAL directly); TimestampWrites is the odd one out.

### Non-blocking follow-up (nice to fix in this PR if easy)

`RenderBundle.Release()` destroys immediately. Prefer DestroyQueue-deferred destroy (ADR-056), same pattern as `ShaderModule` / `Sampler` / `QuerySet`, so a bundle still referenced by an in-flight submission isn’t torn down early. Also guard `b.device == nil` before `halDevice()`.

Happy to re-review as soon as TimestampWrites actually flow through `BeginRenderPass`. Appreciate the careful rebase work — this is close!

gusevgrishaem1:
Hi! Thanks for the detailed review!
I've fixed the `TimestampWrites` path so it now flows through `BeginRenderPass`.
I also addressed the `RenderBundle.Release()` follow-up.
Could you please take another look when you have time?

lkmavi (dismissed):
Hi @gusevgrishaem1 — thanks for the follow-up 🙏

Rechecked the latest commit (`931670d`):

- **TimestampWrites** now flow through the live path: public → `convertRenderPassDesc` → `core.RenderPassDescriptor` → `convertRenderPassDescriptor` → HAL. `TestBeginRenderPassTimestampWrites` asserts what `BeginRenderPass` actually sees on the HAL encoder (absent / nil / released / active).
- **RenderBundle.Release** matches ADR-056: DestroyQueue-deferred destroy, nil-device / nil-hal guards, idempotent. Covered by `TestRenderBundleRelease`.

Earlier rounds look settled as well (RenderBundle API, `ResolveQuerySet`, `ExecuteBundles`, Finish(desc), post-Finish guards, QuerySet kept from #347 + `halQuerySet()`).

**LGTM** — happy to merge once CI is fully settled. `codecov/patch` is still red, but the missing `core/command.go` lines are exercised by the public BeginRenderPass test; treating that as non-blocking.

Thanks for the careful iteration — nice first contribution 🎉

kolkov (changes requested):
Thanks for the careful iteration, @gusevgrishaem1 — the rebase after #347, TimestampWrites fix through the live path, and deferred RenderBundle destruction all look solid.

One blocking issue:

## Draw / DrawIndexed must use struct params (ADR-072) — blocking

Since v0.34.0 all public draw methods take struct params. `RenderPassEncoder.Draw(args gputypes.DrawArgs)` and the HAL interface (`hal.RenderBundleEncoder.Draw(args gputypes.DrawArgs)`) both use `gputypes.DrawArgs` / `gputypes.DrawIndexedArgs`.

`RenderBundleEncoder.Draw(vertexCount, instanceCount, firstVertex, firstInstance uint32)` breaks this convention — the public API inconsistency will be visible in pkg.go.dev.

Fix:

```go
func (e *RenderBundleEncoder) Draw(args gputypes.DrawArgs) {
    if e.finished.Load() {
        return
    }
    e.hal.Draw(args)
}

func (e *RenderBundleEncoder) DrawIndexed(args gputypes.DrawIndexedArgs) {
    if e.finished.Load() {
        return
    }
    e.hal.DrawIndexed(args)
}
```

Tests update accordingly (`e.Draw(3, 1, 0, 0)` → `e.Draw(gputypes.DrawArgs{VertexCount: 3, InstanceCount: 1})`).

## Non-blocking: duplicate `RenderPassTimestampWrites` struct

`RenderPassTimestampWrites` is defined in both `descriptor.go` (public layer) and `core/command.go` (core layer). The `descriptor.go` version feeds `RenderPassDescriptor.toHAL()`, which is a pre-existing converter that no backend calls in production — all three backends (`_native`, `_rust`, `_browser`) use their own converters. The struct and `toHAL()` mapping are not harmful, just redundant with the live path through `convertRenderPassDesc`. Not a regression from this PR — the pattern exists on `main`.

---

Everything else is good: build tags, lifecycle guards, test coverage (496 LOC), DestroyQueue-deferred destroy, `halQuerySet()` accessor, post-Finish encoder guards. Nice work across 5 iterations.

gusevgrishaem1:
Hi! Thanks a lot for the detailed review and feedback!
I’ve made the requested changes and fixed the blocking issue. Could you please take another look when you have time?

kolkov (approved):
Blocking resolved: `Draw()` and `DrawIndexed()` now use `gputypes.DrawArgs` / `gputypes.DrawIndexedArgs` per ADR-072. Tests updated. LGTM — nice first contribution.

PR #354, opened 2026-09-07
ci: harden Dependencies job against go-mod-outdated proxy flakes

lkmavi:
## Summary
- The advisory **Dependencies** job was failing PRs when `go install github.com/psampaz/go-mod-outdated@latest` hit a transient `proxy.golang.org` `INTERNAL_ERROR` (e.g. [#353 run](https://github.com/gogpu/wgpu/actions/runs/34125006674/job/101751523783)).
- Pin `go-mod-outdated@v0.9.0`, retry install up to 3 times, and `continue-on-error` so a module-proxy blip cannot red the check; skip the direct-deps step if install still fails. Ecosystem checks still run.

## Test plan
- [x] CI on this PR: Dependencies job succeeds (or soft-continues) even if a single install attempt flakes
- [x] Confirm direct-deps step runs when install succeeds; ecosystem step always runs
- [x] Re-check that a green PR is no longer blocked solely by this advisory job

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -299,7 +299,7 @@ jobs:
 
   # Dependency freshness check
   # Uses go-mod-outdated (https://github.com/psampaz/go-mod-outdated)
-  # Non-blocking: reports outdated deps as warnings
+  # Non-blocking: reports outdated deps as warnings; never fails the workflow.
   deps:
     name: Dependencies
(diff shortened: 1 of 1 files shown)


kolkov (approved):
CI hardening — pin version, retry with backoff, continue-on-error for advisory check. Standard practice, prevents proxy.golang.org flakes from blocking PRs. LGTM.

